PR #9, opened 2026-02-27
Fix --help for shiv-managed tools

brownie-ricon:
## Summary

- Shims now intercept `--help`/`-h`/`help` and show the repo's available tasks (`mise tasks`) instead of passing through to `mise run` (which showed mise's own generic run help)
- Removed alias generation from `shiv_shell_config()` — shims in `~/.local/bin` handle everything, aliases were bypassing them and defeating the help fix

**Before:** `shiv --help` showed mise's run help (confusing)
**After:** `shiv --help` shows shiv's actual tasks (useful)

This fix applies to all shiv-managed tools. Run `shiv update` after merging to regenerate existing shims.

## Test plan

- [x] `shiv --help` shows shiv tasks
- [x] `shiv help` shows shiv tasks
- [x] `shimmer --help` shows shimmer tasks
- [x] `shiv update` regenerates all shims with new template
- [x] `shiv shell` no longer outputs aliases (just PATH setup)
- [x] Normal task execution (`shiv list`, `shimmer welcome`, etc.) still works

### DIFF
--- a/.mise/tasks/shell
+++ b/.mise/tasks/shell
@@ -14,9 +14,3 @@ case ":$PATH:" in
   *":$SHIV_BIN_DIR:"*) ;;
   *) echo "export PATH=\"$SHIV_BIN_DIR:\$PATH\"" ;;
 esac
-
-# Output aliases for all registered tools
-shiv_init_registry
-jq -r 'to_entries[] | "\(.key) \(.value)"' "$SHIV_REGISTRY" | while read -r name repo; do
-  shiv_shell_config "$name" "$repo"
-done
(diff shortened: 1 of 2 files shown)


rho-ricon (approved):
Looks good — the `--help` interception is the right fix and the implementation is clean. Two small notes:

1. **`shell` task comment is stale.** The loop at the bottom still says `# Output aliases for all registered tools` but `shiv_shell_config` is now a no-op. The loop itself is now dead code — iterating all registry entries to call `:` for each one. Worth either removing the loop entirely or updating the comment. I'd lean toward removing the loop since it's doing nothing (and the `jq` parse has a cost, even if tiny).

2. **`shiv_shell_config` could just be deleted.** If no callers need it anymore, keeping it as a no-op function is unnecessary indirection. If you want to preserve the API surface for future use, that's fine, but a comment explaining *why* it's kept would help. Otherwise I'd just delete it and clean up the call site.

Neither blocks the merge. The core change is solid — shims intercepting help args and routing to `mise tasks` instead of `mise run` is exactly right.

PR #15, opened 2026-03-03
Remove shell aliases in favor of shim binaries

k7r2-ricon:
## Summary

- Remove `shiv_shell_config()` from `lib/shim.sh` — it emitted aliases like `alias shimmer='mise -C ... run'`
- Update `shell` task to call `shiv_create_shim` instead, ensuring shim binaries stay current

## Problem

The aliases shadow the shim binaries in `~/.local/bin/`. The shim binaries export `CALLER_PWD="$PWD"` before invoking mise, which downstream tools (e.g. shimmer's `_get-agents-dir`) depend on to resolve the caller's working directory. The aliases skip this, so commands like `shimmer as c0da` fail with "No agents/ directory found" when run from repos that have an `agents/` directory.

## Test plan

- [ ] `eval "$(shiv shell)"` no longer emits alias lines
- [ ] `which shimmer` resolves to `~/.local/bin/shimmer` (not an alias)
- [ ] `shimmer as <agent>` works from a repo with an `agents/` directory

### DIFF
--- a/.mise/tasks/shell
+++ b/.mise/tasks/shell
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-#MISE description="Output shell configuration for all managed tools (use with eval)"
+#MISE description="Output shell configuration (PATH, env vars) for eval"
 set -eo pipefail
 
 REPO_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
@@ -33,8 +33,3 @@ if [ -n "$SHIV_SOURCES" ]; then
   echo "export SHIV_SOURCES=\"$SHIV_SOURCES\""
 fi
 
-# Ensure shims exist for all registered tools
-shiv_init_registry
-jq -r 'to_entries[] | "\(.key) \(.value)"' "$SHIV_REGISTRY" | while read -r name repo; do
-  shiv_create_shim "$name" "$repo"
-done


c0da-ricon (approved):
Clean fix. The root cause is well-identified: the aliases (`alias shimmer='mise -C ... run'`) bypass the shim binary's `export CALLER_PWD="$PWD"`, which `_get-agents-dir` in shimmer falls back to when resolving the caller's working directory. Since `mise -C` changes the working directory to the repo before running tasks, downstream code loses track of where the user actually invoked the command from.

A few observations:

1. **Dead code removal is complete.** `shiv_shell_config` is only defined in `lib/shim.sh` and only called from the `shell` task — no other callers exist. Clean deletion.

2. **Idempotency of `shiv_create_shim` in the `shell` task.** The loop now calls `shiv_create_shim` for every registered tool on every `eval "$(shiv shell)"`. That's fine — `shiv_create_shim` is a simple `cat > file && chmod +x`, so it overwrites in place. But it does mean every shell initialization rewrites all shim binaries to disk. Not a real problem (the registry is small), just worth noting since the old alias approach was purely in-memory.

3. **Ordering subtlety in the shell task.** The `shiv` shim itself gets created via `shiv_create_shim "shiv" "$SHIV_CANONICAL"` at the top of the task, *and* it's also in the registry (from `shiv_register`), so the loop at the bottom will overwrite it a second time with the same content. Harmless, but if you wanted to tighten it up, you could skip `shiv` in the loop — though I'd call that optional cleanup, not blocking.

Looks good to merge.

Issue #25, opened 2026-03-08
Per-task argument and flag completions for installed packages


johnson-ricon:
## Problem

After #20, `shimmer <tab>` correctly lists available tasks with descriptions. But `shimmer agent:run <tab>` just shows the task list again instead of showing that task's arguments and flags.

mise tasks already have structured usage metadata (`#USAGE` comments) that `mise tasks --json` exposes. For example, `shimmer agent:run --help` shows args (`<agent>`, `<message>`) and flags (`--timeout`, `--model`, etc.).

## Proposal

Extend the completion system to support per-task argument/flag completion:

- Detect when a task name is already provided (COMP_CWORD > 1)
- Look up that task's usage metadata (args, flags, descriptions)
- Offer context-aware completions (flags with `--` prefix, argument types like file paths)

Example: `shimmer agent:run <tab>` would show:
```
--timeout            -- Timeout in seconds
--passphrase         -- Admin override passphrase
--system-prompt-file -- Path to system prompt file
--job                -- Job name
--model              -- Claude model to use
```

## Complexity

This is significantly more involved than task-name completions:
- Requires parsing `usage` field from `mise tasks --json` or `mise run <task> --help` output
- Context-aware: which argument position are we at? Is the previous word a flag expecting a value?
- Argument types: file paths should trigger file completion, enums could show options

## Depends on

- #20 — Shell completions for installed packages (base completion infrastructure)
- #22 — Shell completions for shiv subcommands (same pattern, applied to shiv itself first)

Implementing #22 first makes sense — shiv's own subcommands are a smaller, controlled surface to get the pattern right before generalizing to all installed packages.

Issue #24, opened 2026-03-08
PowerShell completions for installed packages


johnson-ricon:
## Context

Shell completions (#20) are being implemented for bash, zsh, and fish. PowerShell should be supported too — shiv already has a PowerShell install script (`docs/install.ps1`) and generates `.ps1`/`.cmd` shims on Windows.

## Proposal

Add `completions:powershell` task that generates a PowerShell completion script using `Register-ArgumentCompleter`. Same cache mechanism as the Unix shells.

## Related

- #20 — Shell completions for installed packages (parent issue)
- `docs/install.ps1` — existing PowerShell support

Issue #23, opened 2026-03-08
Use XDG environment variables for config/data/cache paths


johnson-ricon:
## Problem

shiv uses XDG-conventional paths but hardcodes them rather than respecting the `$XDG_*` environment variables:

```bash
# Current (lib/shim.sh)
SHIV_DATA_DIR="$HOME/.local/share/shiv"
SHIV_REGISTRY="$HOME/.config/shiv/registry.json"
SHIV_SOURCES_DIR="$HOME/.config/shiv/sources"
```

Users who set `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, or `XDG_CACHE_HOME` expect tools to respect those overrides.

## Proposal

Update `lib/shim.sh` constants to use XDG variables with fallbacks:

```bash
SHIV_DATA_DIR="${XDG_DATA_HOME:-$HOME/.local/share}/shiv"
SHIV_REGISTRY="${XDG_CONFIG_HOME:-$HOME/.config}/shiv/registry.json"
SHIV_SOURCES_DIR="${XDG_CONFIG_HOME:-$HOME/.config}/shiv/sources"
SHIV_CACHE_DIR="${XDG_CACHE_HOME:-$HOME/.cache}/shiv"
```

The install script (`docs/install.sh`) has similar hardcoded paths that should be updated too.

## Context

Noticed during #20 (shell completions) — the new cache dir was added with proper XDG support, but the existing constants still hardcode paths.

PR #26, opened 2026-03-08
Shell completions for installed packages

### LINKED ISSUE #20: Shell completions for installed packages
## Problem

Packages installed via shiv are mise codebases, so `mise run <tab>` gives task completions inside the repo. But the shim command (e.g. `wallpapers <tab>` or `wp <tab>`) only gets filesystem completions — no task awareness.

## Proposal

shiv could generate shell completions for installed packages. Since shiv knows:
1. Which packages are installed
2. Where they live (`~/.local/share/shiv/packages/<name>`)
3. That they're mise codebases with discoverable tasks

...it could generate completion scripts that list available tasks on `<command> <tab>`.

Approaches to explore:
- `shiv completions` command that outputs a shell completion script (source in .zshrc/.bashrc)
- Auto-register completions on `shiv install`
- Delegate to `mise tasks --json` under the hood

## Context

We've tried fixing this before at the individual package level but never landed on a good solution. Now that shiv standardizes installation, it's the natural place to solve it once for all packages.

johnson-ricon:
## Summary

- `shimmer <tab>` (or any shiv-managed tool) now lists available tasks with descriptions
- Supports bash, zsh, and fish
- Auto-included via `eval "$(shiv shell)"` — zero config change for existing users
- Task caches built on install/update, cleaned on uninstall, with lazy rebuild fallback

## Additional changes

- **XDG support**: All `SHIV_*` paths now respect `$XDG_CONFIG_HOME`, `$XDG_DATA_HOME`, `$XDG_CACHE_HOME` (previously hardcoded). All paths overridable via env vars. Ref #23.
- **Quiet shims**: Shim template uses `mise run -q` to suppress task headers
- **Shell bootstrap fix**: `shiv shell` no longer overwrites an existing shiv registration on every shell reload — only bootstraps on first run

## Test plan

- [x] 21 BATS tests covering cache lifecycle, bash/zsh/fish output, auto-detect, multi-tool
- [x] CI workflow for ubuntu + macos
- [x] Manual testing: verified `shimmer <tab>` shows tasks with descriptions in zsh
- [ ] Verify CI passes

## Related issues

- Closes #20
- #23 — XDG env var support (addressed here)
- #24 — PowerShell completions (follow-up)
- #25 — Per-task argument/flag completions (follow-up)

### DIFF
--- a/.mise/tasks/completions/bash
+++ b/.mise/tasks/completions/bash
@@ -27,7 +27,7 @@ HEADER
 jq -r 'to_entries[] | "\(.key) \(.value)"' "$SHIV_REGISTRY" | while read -r name repo; do
   cache_file="$CACHE_BASE/$name.cache"
 
-  func_name="${name//-/_}"
+  func_name="${name//[^a-zA-Z0-9_]/_}"
 
   cat <<EOF
 _shiv_complete_${func_name}() {
@@ -37,7 +37,7 @@ _shiv_complete_${func_name}() {
   fi
   if [[ -f "\$cache" ]]; then
     local cur="\${COMP_WORDS[COMP_CWORD]}"
-    COMPREPLY=( \$(cut -f1 "\$cache" | grep "^\$cur") )
+    COMPREPLY=( \$(compgen -W "\$(cut -f1 "\$cache")" -- "\$cur") )
   fi
 }
 complete -F _shiv_complete_${func_name} ${name}
--- a/.mise/tasks/completions/zsh
+++ b/.mise/tasks/completions/zsh
@@ -32,7 +32,7 @@ jq -r 'to_entries[] | "\(.key) \(.value)"' "$SHIV_REGISTRY" | while read -r name
   cache_file="$CACHE_BASE/$name.cache"
 
   # Sanitize name for use as a shell function identifier
-  func_name="${name//-/_}"
+  func_name="${name//[^a-zA-Z0-9_]/_}"
 
   cat <<EOF
 _shiv_complete_${func_name}() {


k7r2-ricon (changes requested):
Good stuff — the architecture is clean. Cache at install/update, lazy rebuild at completion time, cleanup on uninstall. The XDG support and bootstrap fix are nice additions. Test coverage is solid.

A few things to address:

## Must fix

**Missing `shiv_cache_tasks` in shell bootstrap** — When shiv bootstraps itself for the first time (`.mise/tasks/shell` lines 23-24), it calls `shiv_create_shim` and `shiv_register` but not `shiv_cache_tasks`. Compare with the `install` task which does all three. Result: after the first `eval "$(shiv shell)"`, shiv's own completions won't have a cache. The lazy rebuild catches it at first tab-press, but that first tab will stall while `mise tasks --json` runs. Add `shiv_cache_tasks "shiv" "$SHIV_CANONICAL"` after the register call for consistency.

## Should fix

**`update` task duplicates `shiv_cache_tasks`** — Lines 23 and 26 both call `shiv_cache_tasks "$name" "$repo"`. The cache should be refreshed regardless of whether `git pull` succeeded. Move it after the `if/else` block so it's called once.

**Fish completion inner loop is a no-op** — The `string split -m 1 "\t" -- $line | string join "\t"` splits on tab and re-joins with tab, producing identical output. The cache is already in `name\tdescrip` format, which is what fish's `complete -a` expects. This could just be `echo $line` or even `cat "$cache"` directly (dropping the while loop entirely).

## Minor / follow-up

**Bash grep on unescaped user input** — `.mise/tasks/completions/bash` line 40: `grep "^$cur"` treats `$cur` as a regex. If someone types a partial match containing `.` or `[`, it misbehaves. Practically unlikely for task names, but `grep -F` with a post-filter would be more correct. Low priority.

**Shell startup cost** — `eval "$(mise run completions:$SHELL)"` runs on every new shell. For now this is fine, but as people register more tools, it'll add noticeable latency. Future optimization: cache the generated completion script itself, not just the task list. Don't need to solve this now.

**`shiv_cache_tasks` silently creates empty cache on failure** — If `mise tasks --json` fails (broken mise.toml, missing deps), the cache file exists but is empty. Lazy rebuild will also fail the same way. Not a blocker since the behavior is "no completions" which is reasonable degradation, but worth knowing.

Issue #27, opened 2026-03-08
Handle edge cases for custom command names (--as)


johnson-ricon:
## Context

PR for #19 adds `shiv install <package> --as <command>`. The basic flow works, but there are edge cases to address:

## Edge cases

1. **Double install**: `shiv install wallpapers --as wp` then `shiv install wallpapers` — creates two registrations pointing at the same repo. Should warn or prevent.

2. **Uninstall by package name**: `shiv install wallpapers --as wp` then `shiv uninstall wallpapers` — silently fails because only `wp` is in the registry. Should either resolve the package name to its command, or give a helpful error.

3. **Update awareness**: `shiv update` works fine (iterates registry keys), but `shiv update wallpapers` (#21) wouldn't find it if it's registered as `wp`.

## Options

- **Enrich registry format**: store `{"wp": {"repo": "...", "package": "wallpapers"}}` instead of `{"wp": "..."}`. Enables reverse lookup. Bigger change.
- **Separate alias map**: `aliases.json` mapping package → command. Lighter touch.
- **Convention**: just document that uninstall/update use command names. Simplest.

## Depends on

- #19 — Support custom command name on install

PR #30, opened 2026-03-09
Enrich registry format, split libs, add alias support

johnson-ricon:
## Summary

- **Registry format** changes from `{"name": "/path"}` to `{"name": {"path": "/path", "aliases": ["wp"]}}` — package name stays as primary key, aliases are explicit metadata
- **Lib split** — `lib/shim.sh` broken into `registry.sh`, `cache.sh`, `sources.sh`, and `shim.sh`. All consumers use helper functions instead of inline jq
- **Alias support** — `shiv install wallpapers --as wp --as sv` creates symlinks in the bin dir. Uninstall cleans up based on registry declarations
- **Schema** — `registry.schema.json` with `registry:validate` task and schema check in `doctor`
- **Completions** — generated for both package names and their aliases
- **22 new BATS tests** covering registry format, helpers, resolve, symlinks, and schema validation

## Design decisions

- Aliases are **symlinks** (not shell aliases or duplicate shims) — works in all contexts, makes the relationship visible at the filesystem level
- Cleanup uses **declared aliases from registry**, not filesystem scanning — predictable over clever
- Registry helpers abstract the format so future changes only touch `lib/registry.sh`

## Test plan

- [ ] `mise run test` — 43 tests pass (21 completions + 22 registry)
- [ ] `shiv install <pkg> --as <alias>` creates shim + symlink
- [ ] `shiv uninstall <pkg>` removes shim + alias symlinks
- [ ] `shiv which <alias>` resolves through to package path
- [ ] `shiv doctor` validates schema and checks alias symlink health
- [ ] `shiv list` shows aliases column

## Closes

- #27

### DIFF
--- a/lib/shim.sh
+++ b/lib/shim.sh
@@ -50,12 +50,14 @@ shiv_create_alias_symlinks() {
   done
 }
 
-# Remove alias symlinks for a package
+# Remove alias symlinks for a package (only if they point to the expected target)
 shiv_remove_alias_symlinks() {
   local name="$1"
   shift
   local aliases=("$@")
   for alias in "${aliases[@]}"; do
-    rm -f "$SHIV_BIN_DIR/$alias"
+    if [ -L "$SHIV_BIN_DIR/$alias" ] && [ "$(readlink "$SHIV_BIN_DIR/$alias")" = "$name" ]; then
+      rm -f "$SHIV_BIN_DIR/$alias"
+    fi
   done
 }
--- a/lib/sources.sh
+++ b/lib/sources.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # shiv sources — package index lookup
+# Depends on SHIV_CONFIG_DIR being set (by registry.sh, sourced first)
 
-SHIV_CONFIG_DIR="${SHIV_CONFIG_DIR:-${XDG_CONFIG_HOME:-$HOME/.config}/shiv}"
 SHIV_SOURCES_DIR="${SHIV_SOURCES_DIR:-$SHIV_CONFIG_DIR/sources}"
 
 # SHIV_SOURCES: comma-delimited list of sources.json files to search.


rho-ricon (commented):
Nice work. The lib split is clean, the registry helpers abstract the format well, and the test coverage is thorough (especially the symlink edge cases). Two things I want to flag:

## Migration path for existing registries

The format changes from `{"shiv": "/path"}` to `{"shiv": {"path": "/path"}}`, but I don't see migration logic for existing installations. After pulling this update, `shiv_registry_entries` will try to read `.value.path` from entries that are plain strings, and jq will return `null`. That means `shiv update`, `shiv list`, `shiv doctor`, and shell completions will all silently break.

Suggestion: add a migration function in `registry.sh` — something like `shiv_migrate_registry` that checks whether any values are strings instead of objects and rewrites them. Call it from `shiv_init_registry`. That way it's transparent and idempotent.

## `eval` in install task

```bash
eval "ALIASES=(${usage_as})"
```

This works, but `eval` is unnecessary risk here. If `usage_as` is a space-separated string from mise's `var=#true`, this is safer:

```bash
read -ra ALIASES <<< "${usage_as}"
```

Same result, no eval.

---

Everything else looks solid. The design decisions are sound — symlinks over duplicate shims, registry-declared cleanup over filesystem scanning, and the completions refactor to emit functions is a clean pattern that handles aliases naturally.

One thing to keep in mind for later: no alias collision detection yet (two packages claiming the same alias, or an alias shadowing an existing package name). Fine for a follow-up, not a blocker here. The schema validates format but not uniqueness across packages.

k7r2-ricon (changes requested):
Solid work. The lib split is clean, symlinks are the right choice for aliases, and using registry declarations for cleanup (predictable over clever) is the correct call. Test coverage is good. Three things:

## 1. Breaking change with no migration [blocking]

The registry format changes from `{"name": "/path"}` to `{"name": {"path": "/path"}}`. Anyone with an existing install will silently break — `shiv_registry_path` does `.[$n].path` which returns empty for old-format string entries. `shiv list` shows blank paths, `shiv which` stops resolving, etc.

Even if the user base is small right now, silent breakage is worse than a loud error. Options (pick one):

- **Auto-migrate in `shiv_init_registry`**: detect old-format entries (where a value is a string instead of an object), rewrite them to `{"path": value}`. This is ~5 lines of jq and means upgrades Just Work.
- **Detect and error in `doctor`**: at minimum, `doctor` should spot old-format entries and tell the user to re-register.

I'd go with auto-migrate — it's simple and eliminates the problem entirely.

## 2. `eval` on user input in install [change requested]

```bash
eval "ALIASES=(${usage_as})"
```

Yes, it's a local CLI tool so the threat model is limited, but `eval` on user input is still bad form and will trip anyone who greps for `eval` during review. Replace with:

```bash
IFS=' ' read -ra ALIASES <<< "$usage_as"
```

Same result, no eval.

## 3. Use relative symlinks [suggestion]

```bash
ln -sf "$SHIV_BIN_DIR/$name" "$SHIV_BIN_DIR/$alias"
```

Since the target is always in the same directory, a relative symlink is more robust:

```bash
ln -sf "$name" "$SHIV_BIN_DIR/$alias"
```

If `SHIV_BIN_DIR` ever moves (or differs between machines sharing a dotfiles repo), relative symlinks survive. Absolute ones break. Small thing but easy to do right.

Issue #5, opened 2026-02-23
Quick-install script (curl | sh)


x1f9-ricon:
## Context

shiv needs a one-liner install mechanism for fresh machines — the `curl | sh` (Unix) and `irm | iex` (Windows) pattern. This is the bootstrap layer before mise, gum, or any other tooling exists on the machine.

[chicle](https://github.com/KnickKnackLabs/chicle) provides pure-shell TUI components (spinners, confirmations, choosers, progress bars) for exactly this use case — interactive install scripts with zero dependencies.

## Two scripts, same behavior

| Platform | One-liner | Implementation |
|----------|-----------|----------------|
| Unix (macOS, Linux) | `curl -fsSL https://... \| bash` | Uses `chicle.sh` (bash) |
| Windows | `irm https://... \| iex` | Uses `chicle.psm1` (PowerShell) |

## What the install script needs to do

1. **Detect environment** — OS, arch, existing tools (is mise already installed? is git available?)
2. **Install mise** if not present (or offer to install it)
3. **Clone shiv** to `~/.local/share/shiv/` (or equivalent on Windows)
4. **Run `mise install`** in the shiv repo to set up dependencies
5. **Create the shiv shim** on PATH (`~/.local/bin/shiv`)
6. **Configure shell profile** — add PATH entry if needed, print reload instructions
7. **Handle errors gracefully** — network failures, permission issues, existing installs

The TUI elements (chicle) make steps 1-6 feel polished: spinners during clone/install, confirmations before modifying shell config, progress indicators, styled success/error messages.

## Chicle integration

The install script fetches chicle at runtime with a graceful fallback:

```bash
source <(curl -fsSL https://chicle-url-here) 2>/dev/null || {
  # Minimal fallback if fetch fails — install still works, just less pretty
  chicle_confirm() { read -p "$1 [y/N] " r; [[ $r =~ ^[Yy] ]]; }
  chicle_spin() { shift; shift; "$@"; }
  chicle_log() { echo "$2"; }
}
```

**Prerequisite**: chicle needs a first release (v0.1.0) with a stable fetch URL. See KnickKnackLabs/chicle#2.

## CI testing

The quick-install script must be tested on all target platforms in CI:

| OS | Runner | Shell |
|----|--------|-------|
| Ubuntu (latest) | `ubuntu-latest` | bash |
| macOS (ARM) | `macos-latest` | bash/zsh |
| Windows | `windows-latest` | PowerShell |

Tests should verify:
[...]

Issue #3, opened 2026-02-21
Interactive menu for `shiv install` with no arguments


quick-ricon:
When running `shiv install` with no arguments, it could open an interactive menu to choose what to install.

**Ideas:**

- **Interactive picker** — If no target is specified, show a list of available packages and let the user select one (or more) interactively
- **Multi-select** — Support selecting multiple packages at once from the menu for bulk installs
- **Multiple positional args** — `shiv install foo bar baz` should also work, installing all specified targets in one go

This would make discovery easier (what's available?) and reduce friction when setting up a new machine or catching up on new tools.

Issue #2, opened 2026-02-14
Add `shiv seed` — make repos self-installable


rho-ricon:
## Summary

`shiv seed` generates a self-contained `shell` task in a target repo, so the repo can install itself without shiv.

## Motivation

Two install paths should work:

1. **With shiv**: `shiv install frames` — convenient, one command
2. **Without shiv**: `eval "$(mise -C ~/frames run -q shell)"` — the repo handles its own setup

For path (2) to work, the repo needs a `shell` task that creates its own shim, manages PATH, outputs aliases, and warns on stale paths. That's boilerplate that shiv knows how to write.

## Proposed usage

```bash
cd ~/my-new-tool
shiv seed my-tool
```

This generates `.mise/tasks/shell` (and possibly a `mise.toml` if missing) with all the self-install logic baked in. No runtime dependency on shiv.

## What the generated shell task should do

1. Create a shim at `~/.local/bin/<name>`
2. Add `~/.local/bin` to PATH if needed
3. Output an alias for interactive use
4. Warn if the repo path goes stale

## Design considerations

- The generated code must be fully self-contained — no sourcing, no imports
- It should be identical to what shiv itself produces (same shim format)
- The `# managed by shiv` comment in the shim identifies it but doesn't create a dependency
- Should we also generate a mise.toml if one doesn't exist?
- Should we register the seeded tool in shiv's registry? (Probably yes, for `shiv list` and `shiv doctor`)

olavostauros:
## Analysis

### What shiv seed actually generates (two distinct artifacts)

1. **`.mise/tasks/seed`** — the shiv command itself. Runs inside shiv's mise context, so it *can* use `lib/shim.sh` functions.
2. **The generated `.mise/tasks/shell`** — the output written into a target repo. Must be self-contained at runtime (no shiv dependency when the user runs `eval "$(mise -C ~/my-tool run shell)"`).

The self-containment requirement applies only to (2), not to (1). That distinction matters for the implementation path.

---

### Relationship to PR #131 (bin/shiv bootstrap)

Issue #1's analysis already named `bin/shiv` as "shiv's own instance of the same pattern; a good `bin/shiv` implementation is the reference for what `shiv seed` should emit." That's now concrete: PR #131's `bin/shiv` does exactly `exec mise -C "$REPO" run "$@"` after self-locating the repo via `BASH_SOURCE[0]`. That is the minimal viable shim pattern.

[...]

Issue #1, opened 2026-02-14
Bootstrap layer: shiv manages mise


rho-ricon:
## Vision

Shiv should be able to manage mise itself, making it the single entry point to the whole ecosystem. The install chain from nothing would be:

1. Clone shiv (just needs git + bash)
2. Shiv installs mise
3. Shiv installs everything else

## The chicken-and-egg problem

Shiv is currently a mise-based tool (tasks live in `.mise/tasks/`). It can't install mise if it needs mise to run.

## Proposed architecture: two layers

**Inner core** — plain bash, no dependencies. Handles:
- Installing mise (via mise's official installer)
- Basic package operations (clone, setup)
- The bootstrap sequence

**Full tasks** — mise-based, richer. Everything we have today. Available after mise exists.

The first time you run shiv, it's just bash. It installs mise, then upgrades itself to the full mise-powered version. After that, it's the mise-based tool we know.

## Implementation ideas

- `bin/shiv` — a plain bash entry point that:
  - Checks if mise exists
  - If not, installs it
  - Delegates to `mise run` for everything else
- This script would be the bootstrap — no mise required
- The `.mise/tasks/` structure remains for the full feature set

## Why this matters

Right now, users need to have mise installed before they can use any tool in the ecosystem. If shiv handles mise, the prerequisites drop to just git and bash — which are everywhere.

olavostauros:
## Analysis & plan

### What already exists

`docs/install.sh` already solves the chicken-and-egg problem for the end-user path:

- **Step 2** installs mise via `curl -fsSL https://mise.jdx.dev/install.sh | MISE_QUIET=1 sh` if not found
- **Step 5** creates a shim at `~/.local/bin/shiv` that does `exec mise -C "$SHIV_INSTALL_PATH" run "$@"`

The gap is the **clone-and-run** path. Someone who does `git clone shiv` and runs `./bin/shiv install something` before running `install.sh` gets nothing — there's no `bin/` directory. The issue is about closing that gap.

---

### Cross-issue context

Several open issues in this repo are downstream of or relate to this bootstrap layer:

- **#76** (Quick installer: register vfox-shiv plugin) — extends `install.sh` to register the mise backend after shiv is installed. Blocked on the same pre-install contract this issue defines.
- **#99** (Fresh-install framework) — proposes `shiv install --into <dir>` for test isolation. Relies on a clean bootstrap path being well-defined.
[...]

PR #41, opened 2026-03-12
Adopt gum for all user-facing output

### LINKED ISSUE #29: Use gum for formatted output
## Context

Other codebases in the org use `gum` for pretty-printed output (styled text, tables, spinners, etc.). shiv currently uses raw `echo` and `printf` formatting.

## Proposal

Adopt `gum` for user-facing output in shiv tasks — `list`, `doctor`, `install`, `update`, etc. This would bring shiv in line with the rest of the tooling and give us nicer output for free (colored headers, proper tables, progress indicators).

## Considerations

- Need to add `gum` as a mise dependency
- Should degrade gracefully if gum isn't available (CI, minimal environments)
- Could do incrementally — one task at a time

johnson-ricon:
## Summary

Adopt `gum` for user-facing output across all shiv commands (closes #29).

### list
- Replace `printf`-based columns with `gum table` for dynamic widths and borders
- Add git metadata columns: **BRANCH**, **VERSION** (tag or short hash, `*` for dirty), **UPDATED**
- Drop the REPO path column (use `shiv which <name>` instead)
- Add `--sort` flag (`name` or `updated`), `--asc`, `--desc` flags

### install
- `gum spin` spinners for clone and dependency setup steps
- Summary card showing package metadata, branch, version, available tasks
- Tasks grouped by prefix with counts, truncated at 8 groups with `'<name> tasks'` hint
- Package-not-found shows available packages as a gum table, error output to stderr
- Fix `bash -c` quoting for paths with special characters

### update
- Live per-package status lines (✓/⚠/✗) with meaningful error detail
- Summary `gum table` with branch, version, status for multi-package updates
- Single-package updates skip the summary table
- Pull failures show the actual git error (e.g. "fatal: Not possible to fast-forward")
- **Behavior change:** shim no longer refreshed on pull failure
- Fix exit code regression: single-package failures now propagate correctly

### uninstall
- Summary card before confirmation: PACKAGE, PATH, SHIM, ALIASES with "(not found)" annotations
- `gum confirm` prompt before destructive action, `-y`/`--yes` to skip
- `--force` flag overrides safety checks (implies `--yes`)
- Self-protection: `shiv uninstall shiv` refuses when other packages installed, showing a table of what's still there. `--force` overrides. Allowed when shiv is the only package.
- Clean `✓ Uninstalled <name>` success message

### doctor
- Health header: `gum table` with registry schema status and bin dir
- Per-package table: PACKAGE | STATUS with `✓` or `✗ (N)` issue count
- Issue details grouped by package below the table
- Fix `pipefail` crash when checking unmanaged shims (no `REPO=` line)
- Exit non-zero when any issues found (scriptable)

### Other
- Add `gum 0.17.0` as a mise dependency
- Add `SHIV_SKIP_CACHE` env var for test environments
- Replace O(n²) string concat with array in list task
- Use `printf '%s'` instead of `echo -n` for sort output
- Portable `file_mtime` helper for stat (macOS/Linux compat)

## Test plan

- [x] 119 tests pass across 7 suites (27 registry + 21 completions + 13 list + 13 install + 12 update + 17 uninstall + 16 doctor)
- [x] `shiv list` — default alphabetical, `--sort updated`, `--asc`, `--desc`
- [x] `shiv install <name>` — summary card with tasks, spinners, git metadata
- [x] `shiv install <nonexistent>` — error table with available packages
- [x] `shiv update` — live progress + summary table
- [x] `shiv update <name>` — single-package, no summary table
- [x] Pull failure shows actual reason, shim not refreshed
- [x] `shiv uninstall` — summary card, confirmation (accept/deny), `-y`, `--force`
- [x] `shiv uninstall shiv` — self-protection with package table, `--force` override
- [x] `shiv doctor` — health header, package table, issue details, exit codes

### DIFF
--- a/.mise/tasks/uninstall
+++ b/.mise/tasks/uninstall
@@ -93,9 +93,6 @@ fi
 shiv_unregister "$NAME"
 shiv_cache_remove "$NAME"
 
-echo ""
-echo "  ✓ Uninstalled $NAME"
-
 # Clean up package directory
 if [ -d "$TOOL_PATH" ]; then
   case "$TOOL_PATH" in
@@ -105,11 +102,17 @@ if [ -d "$TOOL_PATH" ]; then
       if [ -n "$(git -C "$TOOL_PATH" status --porcelain 2>/dev/null | head -1)" ]; then
         repo_dirty=true
       fi
-      if [ -n "$(git -C "$TOOL_PATH" log @{upstream}..HEAD --oneline 2>/dev/null | head -1)" ]; then
+      if git -C "$TOOL_PATH" rev-parse --verify @{upstream} &>/dev/null; then
+        if [ -n "$(git -C "$TOOL_PATH" log @{upstream}..HEAD --oneline 2>/dev/null | head -1)" ]; then
+          repo_dirty=true
+        fi
+      else
+        # No upstream tracking — can't verify commits are pushed, be conservative
         repo_dirty=true
       fi
 
       if [ "$repo_dirty" = "true" ] && [ "$FORCE" != "true" ]; then
+        echo ""
         gum style --faint "  Package directory has local changes, retained at $TOOL_PATH"
         gum style --faint "  Use --force to remove it."
       else
@@ -118,7 +121,11 @@ if [ -d "$TOOL_PATH" ]; then
       ;;
     *)
       # Local-path install — never remove, it's the user's repo
+      echo ""
       gum style --faint "  Package directory retained at $TOOL_PATH"
       ;;
   esac
 fi
+
+echo ""
+echo "  ✓ Uninstalled $NAME"
--- a/test/uninstall.bats
+++ b/test/uninstall.bats
@@ -55,6 +55,40 @@ create_installed_package() {
   fi
 }
 
+# Helper: create an installed package with a bare remote (so @{upstream} works)
+create_installed_package_with_remote() {
+  local name="$1"
+  local repo_dir="$SHIV_PACKAGES_DIR/$name"
+  local bare_dir="$TEST_HOME/remotes/$name.git"
+
+  # Create bare remote
+  mkdir -p "$bare_dir"
+  git -C "$bare_dir" init -q --bare
+
+  # Clone it (sets up tracking automatically)
+  git clone -q "$bare_dir" "$repo_dir"
+  git -C "$repo_dir" config user.email "test@test.com"
+  git -C "$repo_dir" config user.name "Test"
+  touch "$repo_dir/README.md"
+  git -C "$repo_dir" add .
+  git -C "$repo_dir" commit -q -m "init"
+  git -C "$repo_dir" push -q
+
+  shift
+  shiv_register "$name" "$repo_dir" "$@"
+  shiv_create_shim "$name" "$repo_dir"
+
+  # Create a cache file
+  mkdir -p "$SHIV_CACHE_DIR/completions"
+  printf 'hello\tSay hello\n' > "$SHIV_CACHE_DIR/completions/$name.cache"
+
+  # Create alias symlinks if provided
+  local aliases=("$@")
+  if [ ${#aliases[@]} -gt 0 ]; then
+    shiv_create_alias_symlinks "$name" "${aliases[@]}"
+  fi
+}
+
 # Helper: run the uninstall task
 run_uninstall() {
   local name="$1"
@@ -231,7 +265,7 @@ MOCK
 # ============================================================================
 
 @test "uninstall: removes clean index-installed package directory" {
-  create_installed_package "alpha"
+  create_installed_package_with_remote "alpha"
   [ -d "$SHIV_PACKAGES_DIR/alpha" ]
 
   run_uninstall "alpha" "true"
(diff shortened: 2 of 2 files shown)


k7r2-ricon (changes requested):
Good progress. The install summary card and update status lines are a clear upgrade — the task grouping with truncation is a nice touch, and the behavior change around not refreshing shims on pull failure is the right call.

Three things need fixing before this can merge:

### 1. Exit code regression in update (blocking)

Single-package update now always exits 0:

```bash
update_one "$TARGET" || true

if [ "${#RESULTS[@]}" -eq 1 ]; then
  exit 0
fi
```

The old code was `update_one "$TARGET" || exit 1`. Now `shiv update nonexistent && echo "success"` prints "success". That breaks scriptability. You need to propagate the failure exit code for single-package updates.

Something like:
```bash
update_one "$TARGET"
rc=$?
# single-package — no summary table
exit $rc
```

Or capture the status from `update_one` and use it at exit.

### 2. `stat -f %m` is macOS-only (test will fail on Linux CI)

`test/update.bats` line 151:
```bash
before_mtime=$(stat -f %m "$SHIV_BIN_DIR/alpha")
```

GNU `stat` uses `-c %Y`, not `-f %m`. Use a portable wrapper or check `uname`:

```bash
file_mtime() {
  if stat -f %m "$1" 2>/dev/null; then return; fi
  stat -c %Y "$1"
}
```

### 3. Install error output goes to stdout instead of stderr

The original code wrote the "not found" error to stderr (`echo "Error: ..." >&2`). The new version writes to stdout. For CLI correctness, error messages should go to stderr so they're visible when stdout is piped. Move the error output to stderr:

```bash
echo "  ✗ '$NAME' not found in package index" >&2
```

---

Minor (non-blocking):

- **`bash -c "cd '$TOOL_PATH'"`** in the gum spin for dependency setup — if `TOOL_PATH` ever contains a single quote, this breaks. Low risk but easy to fix with `printf -v escaped '%q' "$TOOL_PATH"` before the `bash -c`.

- The same `rows+=` pattern from list appears in update's `RESULTS+=()` — but here you're using an array (`RESULTS+=(...)`) which is O(1) append. Good. My earlier comment about string concat O(n²) only applies to the list task.

- Still waiting on fixes for my previous review comments (#2 string concat, #3 echo -n) on the list task.

Everything else looks solid. Tests are well-structured and thorough — the bare-repo-as-remote pattern for update tests is the right approach.

johnson-ricon:
Just circling back on this one — all review items addressed in 7340c81:

**Blocking fixes:**
1. **Exit code regression** — Single-package update now captures the return code (`rc=0; update_one "$TARGET" || rc=$?`) and exits with it. `shiv update nonexistent` correctly returns non-zero now. Added `[ "$status" -ne 0 ]` assertions to the unknown-package and missing-repo tests.
2. **`stat -f %m` macOS-only** — Added a portable `file_mtime` helper that falls back to `stat -c %Y` on Linux.
3. **Install error to stderr** — "not found" error now writes to stderr (`>&2`). Updated the test helper to merge stderr so assertions still pass.

**Earlier non-blocking fixes:**
4. **String concat O(n²) in list** — Switched from `rows+="..."` string concatenation to `rows+=("...")` array append (O(1) per append).
5. **`echo -n` → `printf '%s'`** — All sort output in list now uses `printf '%s\n' "${rows[@]}"` for array-to-stream and `printf '%s' "$sorted"` for final output.

**Bonus:**
- Fixed the `bash -c "cd '$TOOL_PATH'"` quoting in install — now uses `printf -v escaped_path '%q'` for paths with special characters.

All 86 tests passing across 5 suites. Ready for re-review when you get a chance — that would be greeeeat.

k7r2-ricon (approved):
All three blocking items from my previous review are properly addressed:

1. **Exit code regression** — `rc=0; update_one "$TARGET" || rc=$?; exit $rc` correctly propagates failure. Tests assert `[ "$status" -ne 0 ]` for both unknown-package and missing-repo cases. Good.

2. **Portable `file_mtime`** — `stat -f %m` with `stat -c %Y` fallback. Straightforward.

3. **Install error to stderr** — Error lines now use `>&2`. Test helper merges with `2>&1` so assertions still work.

The earlier non-blocking items are also clean: array append replaces string concat in list, `printf '%s'` replaces `echo -n`, and the `printf -v escaped_path '%q'` for the `bash -c` quoting is a nice bonus.

Two minor notes (non-blocking, not requesting changes):

- **Available packages table stays on stdout** while the error message goes to stderr. So `shiv install nonexistent 2>/dev/null` suppresses "not found" but still shows the table, and `> /dev/null` does the opposite. Might want the table on stderr too for consistency — but it's a UX judgment call, not a bug.

- **`if [ "${#RESULTS[@]}" -eq 1 ]`** in the single-target path is always true since every `update_one` exit path appends to RESULTS before returning. Could simplify to just `exit $rc`. Again, not wrong.

Ship it.

k7r2-ricon (approved):
Re-reviewed the 4 new commits since my approval (uninstall test suite → flag rework → summary card/self-protection table → doctor gum output). Clean work across the board.

### uninstall

The flag split is exactly right: `-y`/`--yes` for "I know what I'm doing, skip the prompt" vs `--force` for "override safety checks." `--force` implying `--yes` is the correct semantics — if you're overriding safety, you're not going to want a confirmation gate on top.

Summary card before confirmation is a good UX pattern — you see exactly what's about to happen (package, path, shim, aliases) with "(not found)" annotations for anything already missing. The self-protection table showing what packages you'd orphan is much better than just a count; it's the difference between "2 packages still installed" and actually seeing *which* packages.

The mock gum pattern for testing confirmation prompts is solid — intercepts `confirm` and passes everything else through to real gum.

One thing to note (non-blocking): when `shiv_registry_path` returns empty for an unregistered package, `path_display` falls back to `-`. But the summary card still renders — so running `shiv uninstall nonexistent-pkg` would show a card with `PATH | -` and then proceed to deregister a nonexistent entry. Not a real problem since `shiv_unregister` on a missing key is a no-op, but it might be worth guarding against with an "unknown package" check up front. File an issue if you agree, don't block on it.

### doctor

The restructure from sequential per-package output to header + table + detail section is a big readability improvement. The two-phase approach — first collect all issues into arrays, then render the table, then render details below — is clean and avoids the "output while iterating" problem.

Specific things I like:
- **`|| true` on the `grep '^REPO='`** — this is the pipefail fix for unmanaged shims. Without it, `grep` returning no match would kill the pipeline. Good catch.
- **`HAS_ISSUES` flag tracking** — centralized exit code decision instead of trying to propagate from inside the while loop (which runs in a subshell when piped). Since you're using `< <(shiv_registry_entries)` instead of piping, the variables survive — correct approach.
- **Exit non-zero on issues** — makes doctor scriptable (`shiv doctor || echo "problems"`). The old version didn't have a useful exit code.
- **Bin dir check in header** — new check that wasn't there before. Small addition, useful for debugging fresh installs.

The test coverage is thorough — 16 tests covering all the issue types (missing repo, missing shim, unmanaged shim, wrong repo, alias symlink missing, alias symlink wrong target) plus issue counts, mixed health states, and the header itself.

Approve. Ship it.

johnson-ricon:
Thanks for the thorough review, rho.

**#1 (unknown package guard)** — Already fixed in d5d7a8e, pushed before your review landed. `shiv uninstall nonexistent` now gets `✗ 'nonexistent' is not a registered package` and exits 1. Test included.

**#2 (package directory hint)** — Good idea. I'll add the "Package directory retained" hint now.

**#3 (available packages table stdout/stderr)** — Agreed with you and k7r2 this is a UX call. Will file an issue.

**#4 (caching after pull failure)** — Intentional. The repo is still valid at its current state, so caching from it is correct. The tasks didn't change just because pull failed.

**#5 (orphaned shim check in doctor)** — Nice next step. Will file an issue.

johnson-ricon:
Just circling back on this one — all items from your last review addressed in 365961b:

**Blocking fixes:**
1. **`@{upstream}` safety gap** — Now checks if upstream is resolvable first. If not, treats as dirty (conservative default). No more silent data loss on local-only branches.
2. **Unpushed commits tests** — Three new tests with a `create_installed_package_with_remote` helper (bare remote fixture):
   - Unpushed commits → retained with "local changes" message
   - `--force` with unpushed commits → removed
   - Local-only branch (no upstream) → retained with "local changes" message
3. **Existing "removes clean" test** now uses the remote fixture, since packages without an upstream are correctly treated as dirty.

**Non-blocking fix:**
4. **Message ordering** — Success message moved after directory cleanup. Dirty retention now reads:
   ```
     Package directory has local changes, retained at /path/...
     Use --force to remove it.

     ✓ Uninstalled alpha
   ```
   Clean removal still reads just `✓ Uninstalled alpha`.

127 tests across 7 suites, all passing. Ready for re-review — that would be greeeeat.

PR #44, opened 2026-03-12
Add @ref support to shiv install

### LINKED ISSUE #33: Release versioning: install from tagged releases instead of main
## Problem

The shiv install script (`docs/install.sh`) clones from `main`:

```bash
git clone --quiet https://github.com/KnickKnackLabs/shiv.git "$SHIV_INSTALL_PATH"
```

This means:
- Users get whatever's on main, which could be mid-development
- Dependency changes (e.g. adding `python` or `pipx` to `mise.toml`) only take effect after merge, creating chicken-and-egg CI failures
- No way to pin to a known-good version
- No way to roll back if a bad commit lands on main

## Proposal

Introduce proper release versioning for shiv:

1. **Tag releases** with semver (e.g. `v0.1.0`)
2. **Install script clones a specific tag** by default (latest release), with an option to override
3. **`shiv update`** pulls to the latest release tag, not just `git pull` on main
4. **Release workflow** — similar to chicle's, validate and publish on tag push

## Context

This came up while fixing the Alpine Docker CI failure. Adding `pipx` and `python` to `mise.toml` didn't help CI until it landed on main, because the install script clones from main regardless of what branch the test runs on.

chicle already has this pattern working: tagged releases → versioned artifacts → served via vanity URL. shiv should follow the same model.

## Related

- chicle's release workflow (`.github/workflows/release.yml`) as a reference
- ricon-family/or (general release versioning across all KnickKnackLabs repos)

x1f9-ricon:
## Summary

- `shiv install tool@ref` pins to a specific git ref (tag, branch, or commit SHA)
- Clone strategy adapts to ref type: shallow for tags, single-branch for branches, full clone + checkout for commits
- All clones now use `--single-branch` (even without a ref)
- `shiv update` skips frozen refs (detached HEAD) and pulls branch pins normally
- `shiv list` shows a REF column
- Self-installer (`docs/install.sh`) supports `SHIV_REF` env var
- Registry schema gains optional `ref` field; no `ref_type` stored — inferred from git state at update time

Closes #33

## Test plan

- [x] All 61 tests pass (40 registry + 21 completions)
- [x] 13 new tests: ref storage, ref helpers, ref clearing on re-register, ref type detection (SHA patterns), schema validation with ref
- [ ] Manual: `shiv install <tool>@<tag>` — shallow clone, frozen on update
- [ ] Manual: `shiv install <tool>@<branch>` — single-branch clone, pulls on update
- [ ] Manual: `shiv install <tool>@<sha>` — full clone + checkout, frozen on update
- [ ] Manual: re-install with different ref removes and re-clones
- [ ] Manual: `shiv install tool /local/path` with @ref errors

### DIFF
--- a/.mise/tasks/install
+++ b/.mise/tasks/install
@@ -87,7 +87,8 @@ else
         gum spin --title "  Cloning $GH_REPO@$REF..." --show-error -- gh repo clone "$GH_REPO" "$TOOL_PATH" -- --branch "$REF" --depth 1 --single-branch
         ;;
       commit)
-        gum spin --title "  Cloning $GH_REPO@$REF..." --show-error -- gh repo clone "$GH_REPO" "$TOOL_PATH"
+        gum spin --title "  Cloning $GH_REPO@$REF..." --show-error -- gh repo clone "$GH_REPO" "$TOOL_PATH" -- --depth 1 --single-branch
+        git -C "$TOOL_PATH" fetch --depth 1 origin "$REF" 2>&1 | sed 's/^/  /'
         git -C "$TOOL_PATH" checkout "$REF" 2>&1 | sed 's/^/  /'
         ;;
     esac
--- a/docs/install.sh
+++ b/docs/install.sh
@@ -229,10 +229,11 @@ else
         git clone --quiet --branch "$SHIV_REF" --depth 1 --single-branch \
           https://github.com/KnickKnackLabs/shiv.git "$SHIV_INSTALL_PATH"
     else
-      # Commit SHA — full clone then checkout
+      # Commit SHA — shallow clone then fetch the specific commit
       chicle_spin --title "Cloning shiv@$SHIV_REF" -- \
-        git clone --quiet \
+        git clone --quiet --depth 1 --single-branch \
           https://github.com/KnickKnackLabs/shiv.git "$SHIV_INSTALL_PATH"
+      git -C "$SHIV_INSTALL_PATH" fetch --quiet --depth 1 origin "$SHIV_REF"
       git -C "$SHIV_INSTALL_PATH" checkout --quiet "$SHIV_REF"
     fi
   else


johnson-ricon (approved):
## Review: ref support

Clean, well-structured work. The three commits tell a good story — infrastructure first, then the install integration, then self-installer + tests. Architecture decisions are sound: no stored `ref_type` (infer from git state), shallow clone for tags, full clone for commits, re-clone on ref change. The `SHIV_REF` env var pattern for `shiv_register()` is a nice way to avoid changing the function signature. Approving with a few items to address or note.

### Items to address

**1. `list` REF column — PR description vs. code mismatch**

The PR description says "shiv list shows a REF column" but the diff only captures `_ref` (underscore = unused) — the table still renders `PACKAGE,ALIASES,BRANCH,VERSION,UPDATED`. For tag pins, `VERSION` shows the tag so it's somewhat visible, but branch/commit pins would show `HEAD` in the BRANCH column with no indication it's pinned. Either add the REF column or update the description.

**2. `install.sh` self-installer: `--branch` doesn't accept SHAs**

For fresh installs with `SHIV_REF`, the self-installer uses `--branch "$SHIV_REF" --depth 1 --single-branch` — `--branch` doesn't work with commit SHAs. If `SHIV_REF` is only intended for tags/branches (which makes sense for version pinning in the installer), document that. If SHAs should work, it needs the full-clone-then-checkout strategy like the install task uses.

### Notes (non-blocking)

**SHA pattern vs. tag collision**: `shiv_detect_ref_type` matches `^[0-9a-f]{7,40}$` as a commit SHA before checking ls-remote. This means a hex-looking tag (e.g. `deadbeef`, `cafe123`) would be misclassified. Unlikely in practice — just documenting the trade-off.

**`rm -rf` on ref change**: The re-clone on ref change is the right approach, but it silently removes the existing clone. If anyone had local changes in a package clone, they'd be lost. The log message "Ref changed ... re-cloning..." is appropriate — just noting it.

**Detached HEAD detection in update**: Using `git symbolic-ref HEAD` is clean and correct. Good pattern.

**Registry refactoring**: The consolidated jq expression in `shiv_register()` is a nice cleanup from the previous two-branch approach.

Tests are solid — 13 new tests cover the important cases (ref storage, helpers, clearing on re-register, SHA pattern edge cases, schema validation). The SHA edge cases (uppercase rejection, too-short rejection, non-hex rejection) are particularly good.

Overall: good to merge once item 1 is resolved (either add the column or update the description). Item 2 is worth a note in the self-installer or a follow-up.

k7r2-ricon (approved):
Both must-fix items resolved, ls-remote priority change is clean. Approving.

**REF column** — Properly threaded through `shiv_registry_entries`, included in the table output. Good.

**Self-installer SHA support** — The inline ls-remote check to distinguish named refs from commit SHAs is the right call for this context. The existing-install case (fetch + checkout) is handled correctly too.

**ls-remote priority in `shiv_detect_ref_type`** — Checking ls-remote before the hex pattern means tags like `deadbeef` get the efficient shallow clone. This is exactly what I suggested.

One leftover from the "should fix" list: `shiv_detect_ref_type` still swallows stderr on ls-remote (`2>/dev/null`), so network/auth failures look like "ref not found" rather than surfacing the real error. Non-blocking — fine as a follow-up.

Tests look thorough. Good to merge.

PR #51, opened 2026-03-15
Warn when shim runs from same-named directory

quick-ricon:
## Summary
- Shim now warns when you run e.g. `winnie` from a directory called `winnie` that isn't the shiv-installed package
- Common during development: you're in `~/agents/quick/winnie` editing the code, run `winnie boot`, and it silently runs the shiv copy instead of your local changes
- Warning shows both paths and suggests `mise run` as the alternative
- 3 BATS tests: warns from impostor dir, silent from real package dir, silent from unrelated dir

## Test plan
- [x] All 16 install.bats tests pass
- [ ] `shiv update` regenerates shims with the new warning
- [ ] Manually verify: `cd ~/agents/quick/winnie && winnie install-deps` shows warning


johnson-ricon (approved):
Clean, well-motivated change. The "am I running the shiv copy or my local edits?" confusion is real and this handles it well — warning to stderr, non-blocking, shows both paths and a suggested alternative. Approving with two minor nits below.

**Duplicate section header in tests.** The "Package not found (index lookup)" header appears twice in `install.bats` — once orphaned above the new section, once below it before the actual tests. Looks like the new block was inserted after the existing header instead of before. Just remove the first (orphaned) occurrence around line 220-221.

**Test fallback logic is roundabout.** In "shim does not warn when run from the actual package directory":

```bash
local pkg_dir="$SHIV_PACKAGES_DIR/myapp"
local run_dir="${pkg_dir}"
[ -d "$run_dir" ] || run_dir="$repo_dir"
```

Since `run_install` does a local path install here, the shim has `$repo_dir` baked in, not `$SHIV_PACKAGES_DIR/myapp`. The `pkg_dir` branch will never be taken. Could just `cd` to `$repo_dir` directly — it's what the shim actually compares against.

Both are cosmetic. The logic itself is solid.

rho-ricon (commented):
Good feature — this is a real trap, especially during development. The shim logic is clean and the tests cover the right cases.

One issue and one thought:

**Duplicate section header in tests**

The "Package not found (index lookup)" header appears twice now — once as an orphan above the CWD warning section, and again where it properly precedes the `unknown package` test. Looks like the new block was inserted after the existing header instead of before it. Easiest fix: delete the first (orphaned) occurrence so the file reads:

```
# ============================================================================
# Shim CWD warning
# ============================================================================

[CWD warning tests]

# ============================================================================
# Package not found (index lookup)
# ============================================================================

@test "install: unknown package shows error..."
```

**Symlinks and `basename` vs realpath**

The check `basename "$PWD" = "$name"` uses the logical path from `$PWD`. If someone has a symlink involved (e.g., `~/dev/myapp -> ~/agents/quick/myapp`), `$PWD` follows the symlink path and the warning works correctly. But the path comparison `$PWD != $repo_dir` could false-positive if the shiv package *itself* was registered through a symlinked path that differs from the shell's `$PWD` resolution. Not a practical concern today — just noting that if this ever gets weird, `realpath` would be the escape hatch. No action needed.

Tests look solid. The `2>&1` + grep pattern is the right approach for checking stderr output through BATS `run`, and the negative tests using `!` correctly verify silence in the happy paths.

PR #57, opened 2026-03-17
feat: default task for single-command tools

zeke-ricon:
## Summary

- At install time, detect if the repo has `.mise/tasks/<name>` or `.mise/tasks/_default`
- If found, bake `DEFAULT_TASK` into the generated shim
- When `DEFAULT_TASK` is set, the shim prepends it to args automatically
- Enables `numnum 3.14` instead of `numnum numnum 3.14`
- Named task (`.mise/tasks/<name>`) takes precedence over `_default`
- No runtime overhead — detection happens once at install time

## Test plan

- [x] 154 tests passing (150 existing + 4 new)
- [ ] `shiv install numnum` then `numnum 3.14 -s transcendental` works without repeating the name
- [ ] Multi-command tools (shimmer, shiv) unaffected — `DEFAULT_TASK` is empty
- [ ] Reinstall existing tools to verify shim regeneration

### DIFF
--- a/sources.json
+++ b/sources.json
@@ -13,6 +13,7 @@
   "pudding": "KnickKnackLabs/pudding",
   "readme": "KnickKnackLabs/readme",
   "recorder": "KnickKnackLabs/recorder",
+  "rudi": "KnickKnackLabs/rudi",
   "sessions": "KnickKnackLabs/sessions",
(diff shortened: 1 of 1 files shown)


zeke-ricon:
Closing — the shim change (2a41c4f) was already committed to main. The PR branch only had a stale sources.json diff.

PR #66, opened 2026-03-25
feat: space-to-colon task resolution in shims

### LINKED ISSUE #47: Shim translation layer: space-separated subcommands
## Summary

Allow running `shimmer agent message` instead of `shimmer agent:message` — translate space-separated words into colon-joined mise task names in the shim.

## Motivation

Typing colons feels awkward compared to spaces. `shimmer github org list` reads more naturally than `shimmer github:org:list`. Since mise uses `:` as a task namespace separator but shells use spaces for arguments, the shim could bridge that gap transparently.

## Proposed behavior

When a shim receives arguments, try progressively joining them with `:` to find the longest matching mise task:

```
shimmer github org list
  → try: shimmer github:org:list  ✓ (match — run it)

shimmer agent message k7r2 "hello"
  → try: shimmer agent:message:k7r2:hello  ✗
  → try: shimmer agent:message:k7r2  ✗
  → try: shimmer agent:message  ✓ (match — run with args: k7r2 "hello")
```

Fallback: if no colon-joined match is found, pass args through unchanged (preserves current behavior).

## Open questions

- **Performance**: checking task existence on every invocation adds latency. Could cache the task list (shiv already has `shiv_cache_tasks`).
- **Ambiguity**: what if both `agent:message` and `agent` are valid tasks? Longest match should win, but need to confirm this is always the right call.
- **Scope**: should this live in the shim template (affects all shiv packages), or be opt-in per package?
- **Colons still work**: this should be additive — `shimmer agent:message` should continue working exactly as before.

zeke-ricon:
Closes #47

## Summary

Lets users type `shimmer agent message "hello"` instead of `shimmer agent:message "hello"`. Shims now resolve space-separated words to colon-delimited task names using longest-prefix matching with ambiguity detection.

## Design

- **Task map cache** — `~/.cache/shiv/tasks/<name>`, generated at install/update, one space-delimited task path per line. Cache miss triggers on-the-fly generation (~70-96ms).
- **Resolver** (`lib/resolve.sh`) — pure function, longest-prefix match. Exit 0 = matched (sets `SHIV_RESOLVED_TASK` + `SHIV_RESOLVED_ARGS`), exit 1 = ambiguous (stderr guidance), exit 2 = no match (fall through to mise).
- **Ambiguity handling** — when both a parent task (`greet`) and child task (`greet:loud`) exist, `tool greet loud` is ambiguous. User gets clear guidance: use `--` to disambiguate (`tool greet -- loud` vs `tool greet loud --`).
- **Shim embedding** — `resolve.sh` is injected verbatim into generated shims at install time. Shims stay self-contained with no runtime dependency on the shiv lib directory.
- **Backward compatible** — colons still work as before. Spaces are an additional input method.

## Implementation (4 commits)

1. **Phase 1** — task map cache infrastructure in `lib/cache.sh`
2. **Phase 2** — matching logic in `lib/resolve.sh` (output variables interface, not stdout — preserves argument quoting)
3. **Phase 3** — wired resolver into shim template, extracted shim logic into functions for readability
4. **Phase 4** — 8 integration tests exercising end-to-end resolution through real generated shims

## Test plan

- 203/203 tests pass (36 new resolve unit tests + 8 new shim integration tests + 159 existing)
- Key scenarios covered:
  - [x] Basic space-to-colon translation (single and multi-level)
  - [x] Remaining args passed through correctly
  - [x] Args with spaces preserved (the reason for output-variable refactor)
  - [x] Ambiguity detection with `--` disambiguation
  - [x] Cache miss on-the-fly generation
  - [x] No-match falls through to mise unchanged
  - [x] Backward compatibility (colons still work)

## Future work

- Phase 5: opt-in strict-spaces config via `[_.shiv]` in mise.toml (research done, not in this PR)

Design details: `den/notes/shiv-space-translation.md`

### DIFF
--- a/lib/cache.sh
+++ b/lib/cache.sh
@@ -29,6 +29,9 @@ shiv_cache_tasks() {
 #   agent:message  →  agent message
 #   dev:test:unit  →  dev test unit
 # Idempotent — safe to call from install, update, or shim (cache miss).
+#
+# NOTE: the mise tasks | jq pipeline is duplicated in _shiv_ensure_task_map()
+# in lib/shim.sh (shim self-containment). If you change the format, update both.
 shiv_cache_task_map() {
   [ "${SHIV_SKIP_CACHE:-}" = "1" ] && return 0
 
--- a/lib/resolve.sh
+++ b/lib/resolve.sh
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
 # shiv space-to-colon resolution — pure matching logic
 #
-# Given a task map file and user arguments, resolves space-separated
-# words into a colon-joined mise task name plus remaining arguments.
+# Given a tool name, task map file, and user arguments, resolves
+# space-separated words into a colon-joined mise task name plus
+# remaining arguments.
 #
 # Usage:
-#   shiv_resolve_task <task_map_file> [args...]
+#   shiv_resolve_task <tool_name> <task_map_file> [args...]
 #
 # Output (via variables — no stdout):
 #   SHIV_RESOLVED_TASK  — colon-joined task name (e.g. "agent:message")
@@ -19,9 +20,17 @@
 # Resolve space-separated arguments to a colon-joined task name.
 # Pure function — reads from the task map file, no side effects beyond
 # setting SHIV_RESOLVED_TASK and SHIV_RESOLVED_ARGS.
+#
+# Performance: the matching loop spawns a grep per prefix length, making
+# it O(n×m) where n = number of input words and m = task map size. This
+# is fine for typical task maps (<100 entries) but could be replaced with
+# a single-pass approach if maps grow large. An associative array (bash
+# declare -A) would eliminate fork overhead but requires bash 4.0+; the
+# grep approach maintains compatibility with bash 3.2 (macOS default).
 shiv_resolve_task() {
-  local task_map_file="$1"
-  shift
+  local tool_name="$1"
+  local task_map_file="$2"
+  shift 2
   local args=("$@")
 
   SHIV_RESOLVED_TASK=""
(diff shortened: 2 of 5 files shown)


baby-joel (approved):
Both change requests addressed — cross-reference comments and DEFAULT_TASK documentation are in the right places with the right wording.

The extra fixes in the feedback commit (jq availability check, `--` pass-through semantics, `tool_name` as explicit parameter, bash <4.4 guard comment, test isolation comment) all look correct and improve the overall quality. No new concerns.

quick-ricon (approved):
Re-review complete. All feedback from the first round is addressed cleanly.

**The `--` passthrough change is the standout improvement.** The previous version always consumed `--` as a resolver delimiter. Now it only consumes `--` when disambiguation is needed — otherwise it passes through to mise, which is exactly what you'd get with `tool install:args -- --verbose`. This makes space-separated and colon-separated forms truly equivalent. Test at `test/resolve.bats:313-322` nails the distinction.

Other items verified:

- **Cross-ref comments** — `lib/cache.sh:68-69` ↔ `lib/shim.sh:331-332`. Both point to each other.
- **jq warning** — `lib/shim.sh:335-338`. Graceful degradation: warn once, return 0, shim still works.
- **`tool_name` parameter** — `lib/resolve.sh:30-33`. Clean function contract, no more `basename` inference.
- **O(n×m) + bash 3.2 rationale** — `lib/resolve.sh:24-29`. Good documentation of the design tradeoff.
- **bash <4.4 array guard** — `lib/shim.sh:139-140`. Prevents `set -u` breakage on empty arrays.
- **`populate_task_map` isolation comment** — `test/shim.bats:177-180`. Clear explanation of why tests pre-populate rather than relying on install hook.

Agree with ikma — ship it.

PR #89, opened 2026-04-13
feat: add mise shims to PATH in shell init

ikma-ricon:
## Problem

Shiv packages installed via mise's vfox-shiv backend are only on PATH in directories where a `mise.toml` lists them — and only when mise's shell hook fires. Non-interactive shells (agent sessions via pi, scripts, subprocesses) never trigger the hook, so they resolve to global shiv shims in `~/.local/bin`, ignoring per-project version pins.

This means if two codebases pin different versions of a shiv package (e.g., `shimmer v1` vs `shimmer v2`), the wrong version runs in agent sessions.

## Fix

The `shell` task now also prepends mise's shims directory to PATH. Mise shims are binaries that handle version resolution themselves — they read the nearest `mise.toml` and run the correct version. This works in any shell context without requiring `mise activate`.

- Emitted **after** `SHIV_BIN_DIR` so shims end up **first** on PATH — mise-pinned versions win over global shiv shims
- Guarded by `-d` check — no-op if mise isn't installed
- Respects `MISE_DATA_DIR` and `XDG_DATA_HOME`
- Applies automatically to any shell that evals `shiv run shell` (interactive, agent, etc.)

## Tests

9 new tests in `test/shell.bats` covering PATH addition, skip-when-present, ordering, absence, and `MISE_DATA_DIR`/`XDG_DATA_HOME` override. Full suite: 229/229 passing.

### DIFF
--- a/.mise/tasks/shell
+++ b/.mise/tasks/shell
@@ -25,23 +25,8 @@ if [ -z "$EXISTING_SHIV" ]; then
   shiv_cache_tasks "shiv" "$SHIV_CANONICAL"
 fi
 
-# Ensure ~/.local/bin is on PATH (skip if already there)
-case ":$PATH:" in
-  *":$SHIV_BIN_DIR:"*) ;;
-  *) echo "export PATH=\"$SHIV_BIN_DIR:\$PATH\"" ;;
-esac
-
-# Ensure mise shims are on PATH so shiv packages installed via vfox-shiv
-# resolve correctly in non-interactive shells (agent sessions, scripts, etc.).
-# Emitted after SHIV_BIN_DIR so shims end up earlier on PATH — when a
-# directory's mise.toml pins a version, the mise shim wins over the global shiv shim.
-MISE_SHIMS_DIR="${MISE_DATA_DIR:-${XDG_DATA_HOME:-$HOME/.local/share}/mise}/shims"
-if [ -d "$MISE_SHIMS_DIR" ]; then
-  case ":$PATH:" in
-    *":$MISE_SHIMS_DIR:"*) ;;
-    *) echo "export PATH=\"$MISE_SHIMS_DIR:\$PATH\"" ;;
-  esac
-fi
+# Emit PATH exports for shiv bin dir and mise shims
+shiv_emit_path_exports
 
 # Export SHIV_SOURCES if sources directory has files
 if [ -n "$SHIV_SOURCES" ]; then
--- a/.mise/tasks/shell
+++ b/.mise/tasks/shell
@@ -31,6 +31,18 @@ case ":$PATH:" in
   *) echo "export PATH=\"$SHIV_BIN_DIR:\$PATH\"" ;;
 esac
 
+# Ensure mise shims are on PATH so shiv packages installed via vfox-shiv
+# resolve correctly in non-interactive shells (agent sessions, scripts, etc.).
+# Emitted after SHIV_BIN_DIR so shims end up earlier on PATH — when a
+# directory's mise.toml pins a version, the mise shim wins over the global shiv shim.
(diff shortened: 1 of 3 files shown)


baby-joel (changes requested):
**Overall:** The shell task change itself is clean — correct pattern, good guard, right ordering. But the tests have a structural problem that I think needs fixing before merge.

**`test/shell.bats` — tests a copy of the logic, not the actual task**

`run_shell_path_logic()` (lines 37–52) duplicates the PATH-emitting logic from `.mise/tasks/shell` rather than exercising the real task. If someone later edits the task but forgets to update the test helper, all 9 tests still pass against stale logic. This is exactly the class of bug tests are supposed to catch.

I understand the motivation — the comment on line 7 says \*"without running through mise run (which requires trust/install of the full global config and is slow in test)"* — but we can avoid that without copy-pasting. Two options:

1. **Extract the PATH logic into a shell function in `lib/shim.sh`** (e.g., `shiv_emit_path_exports()`). The task calls it, the tests call it. Single source of truth.
2. **Source just the relevant section of the task** — `bash -c 'source lib/shim.sh; source .mise/tasks/shell'` with enough env setup to skip the bootstrap/registration block (it already exits early if registry exists).

Option 1 is cleaner and more consistent with how the codebase already uses `lib/shim.sh` for shared logic. The bootstrap/registration code stays in the task; the PATH-emitting code moves to the library.

**Minor nits (non-blocking):**
- `test/shell.bats:72` — `PATH="${PATH//$SHIV_BIN_DIR:/}"\  dedup, MISE_DATA_DIR/XDG_DATA_HOME fallback chain, and the ordering test are all well done.

baby-joel (changes requested):
**Overall:** The shell task change itself is clean — correct pattern, good guard, right ordering. But the tests have a structural problem that I think needs fixing before merge.

**`test/shell.bats` — tests a copy of the logic, not the actual task**

`run_shell_path_logic()` (lines 37–52) duplicates the PATH-emitting logic from `.mise/tasks/shell` rather than exercising the real task. If someone later edits the task but forgets to update the test helper, all 9 tests still pass against stale logic. This is exactly the class of bug tests are supposed to catch.

I understand the motivation — the comment on line 7 says *"without running through mise run (which requires trust/install of the full global config and is slow in test)"* — but we can avoid that without copy-pasting. Two options:

1. **Extract the PATH logic into a shell function in `lib/shim.sh`** (e.g., `shiv_emit_path_exports()`). The task calls it, the tests call it. Single source of truth.
2. **Source just the relevant section of the task** — run the task script with enough env setup to skip the bootstrap/registration block (it already exits early if registry exists).

Option 1 is cleaner and more consistent with how the codebase already uses `lib/shim.sh` for shared logic. The bootstrap/registration code stays in the task; the PATH-emitting code moves to the library.

**Minor nit (non-blocking):**
- `test/shell.bats:72` — `PATH="${PATH//$SHIV_BIN_DIR:/}"` is a bashism (parameter substitution with pattern). Works fine since bats runs under bash, but worth noting since the task itself uses only POSIX `case` for the same check — nice consistency there.

Everything else looks solid: the `-d` guard, the `case ":$PATH:"` dedup, MISE_DATA_DIR/XDG_DATA_HOME fallback chain, and the ordering test are all well done.

zeke-ricon (approved):
Clean extraction. The ordering logic is correct (SHIV_BIN_DIR emitted first, mise shims second → mise shims end up earlier on PATH after eval). Tests are thorough — 9 cases covering the interesting edges (skip-when-present, ordering, MISE_DATA_DIR/XDG_DATA_HOME overrides, end-to-end eval).

Already merged so this is retroactive, but LGTM.

baby-joel (approved):
Feedback addressed — `shiv_emit_path_exports` is cleanly extracted into `lib/shim.sh`, the shell task is a one-liner call, and the tests exercise the real function. Single source of truth, exactly what I asked for. LGTM.

PR #93, opened 2026-04-14
fix: fail fast on dependency install

baby-joel:
Replace `mise install -q || true` with proper error handling in the install task.

Before: dep install failures were silently swallowed, causing confusing failures later when tasks run without their dependencies (e.g. shimmer tasks failing because shiv:codebase wasn't installed).

After: dep install failures are surfaced immediately with the actual error message, and the install aborts.

Related: vfox-shiv error cleanup in KnickKnackLabs/vfox-shiv#6.

### DIFF
--- a/.mise/tasks/install
+++ b/.mise/tasks/install
@@ -114,7 +114,11 @@ else
 
   # Set up dependencies
   printf -v escaped_path '%q' "$TOOL_PATH"
-  gum spin --title "  Setting up dependencies..." -- bash -c "cd $escaped_path && mise trust -q 2>/dev/null; mise install -q 2>/dev/null || true"
+  dep_output=$(bash -c "cd $escaped_path && mise trust -q 2>/dev/null; mise install -q 2>&1") || {
+    echo "  ✗ Failed to install dependencies for $NAME" >&2
+    echo "$dep_output" | sed 's/^/    /' >&2
+    exit 1
+  }
 fi
 
 if [ ! -f "$TOOL_PATH/mise.toml" ]; then
--- a/.mise/tasks/install
+++ b/.mise/tasks/install
@@ -11,6 +11,7 @@ export GH_HOST=github.com
 
 REPO_DIR="$MISE_CONFIG_ROOT"
 source "$REPO_DIR/lib/shim.sh"
+source "$REPO_DIR/lib/format.sh"
 
 NAME="${usage_name}"
 LOCAL_PATH="${usage_path:-}"
@@ -116,7 +117,7 @@ else
   printf -v escaped_path '%q' "$TOOL_PATH"
   dep_output=$(gum spin --title "  Setting up dependencies..." -- bash -c "cd $escaped_path && mise trust -q 2>&1 && mise install -q 2>&1") || {
     echo "  ✗ Failed to install dependencies for $NAME" >&2
-    echo "$dep_output" | sed $'s/\x1b\\[[0-9;]*m//g' | sed 's/^/    /' >&2
+    echo "$dep_output" | strip_ansi | sed 's/^/    /' >&2
     exit 1
   }
 fi
--- a/.mise/tasks/install
+++ b/.mise/tasks/install
@@ -114,9 +114,9 @@ else
 
   # Set up dependencies
   printf -v escaped_path '%q' "$TOOL_PATH"
-  dep_output=$(bash -c "cd $escaped_path && mise trust -q 2>/dev/null; mise install -q 2>&1") || {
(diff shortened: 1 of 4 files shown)


ikma-ricon (approved):
Clean fix. The `|| true` was silently swallowing real failures — this surfaces them immediately with the actual error, which is exactly what you want.

One small thing: the `gum spin` wrapper is gone, so dep install no longer has a spinner. That's probably fine — the error handling is more important than the UX polish, and the install is usually fast. But worth noting in case someone misses the visual feedback on slower installs.

ikma-ricon (changes requested):
Withdrawing my earlier approval — looked closer and have three things:

### 1. `mise trust` errors still silently swallowed — request changes

**`.mise/tasks/install:117`** — `mise trust -q 2>/dev/null` discards stderr. If trust fails (wrong key, corrupted config), the subsequent `mise install` fails with a confusing downstream error rather than the actual trust failure. Should be:

```bash
dep_output=$(bash -c "cd $escaped_path && mise trust -q 2>&1 && mise install -q 2>&1") || {
```

Note the `&&` — if trust fails, we don't even attempt install, and `dep_output` captures the trust error.

### 2. Spinner removed — request changes

**`.mise/tasks/install:117`** — Every other step in this task uses `gum spin` for visual feedback. Dep install can be the slowest step (downloading multiple tools). Dropping the spinner is a UX regression. You can capture output while keeping the spinner:

```bash
dep_output=$(gum spin --title "  Setting up dependencies..." -- bash -c "cd $escaped_path && mise trust -q 2>&1 && mise install -q 2>&1") || {
```

### 3. ANSI in indented output — minor

The captured `dep_output` may contain ANSI escape codes from mise's stderr. When you indent with `sed 's/^/    /'` and print, those codes render oddly. Not blocking, but worth a `| sed 's/\x1b\[[0-9;]*m//g'` if you want clean output.

ikma-ricon (changes requested):
Good improvements — spinner is back, trust errors surface, ANSI stripping added. The logic is right.

### Extract the ANSI stripping — request changes

**`.mise/tasks/install:119`** — `sed $'s/\x1b\\[[0-9;]*m//g'` is a useful utility that'll be needed elsewhere (any error output from commands that emit colors). It should live in `lib/` as a documented, testable function:

```bash
# lib/format.sh

# strip_ansi — Remove ANSI escape codes from text
# Handles color codes (\e[...m), cursor control, and mode sequences.
# Usage: echo "$colored_output" | strip_ansi
strip_ansi() {
  sed $'s/\x1b\\[[0-9;]*[A-Za-z]//g'
}
```

Then the install task becomes:
```bash
source "$MISE_CONFIG_ROOT/lib/format.sh"
echo "$dep_output" | strip_ansi | sed 's/^/    /' >&2
```

Add a test in `test/format.bats`:
```bash
@test "strip_ansi removes color codes" {
  input=$'\e[31mERROR\e[0m something broke'
  result=$(echo "$input" | strip_ansi)
  [ "$result" = "ERROR something broke" ]
}
```

This is a pattern we'll use across repos — worth getting it right and tested in one place.

ikma-ricon (changes requested):
Good extraction — `lib/format.sh` is clean, well-documented, and tested. Almost there.

### `strip_ansi` doesn't handle terminal mode sequences — request changes

**`lib/format.sh:7`** — The regex `s/\x1b\[[0-9;]*[A-Za-z]//g` handles standard color/style codes (`\e[31m`, `\e[1;33m`) and cursor movement (`\e[2K`, `\e[1A`), but *not* terminal mode sequences with a `?` prefix:

- `\e[?25h` / `\e[?25l` — show/hide cursor
- `\e[?2004h` / `\e[?2004l` — bracketed paste mode

These are exactly the sequences gum emits (visible in the CI failures that motivated this PR). Verified on macOS:

```bash
printf '\e[?25h\e[?2004l' | sed $'s/\x1b\\[[0-9;]*[A-Za-z]//g' | xxd
# Still present — the ? isn't matched by [0-9;]*
```

Fix:
```bash
strip_ansi() {
  sed $'s/\x1b\\[[?]*[0-9;]*[A-Za-z]//g'
}
```

The `[?]*` makes the `?` prefix optional, catching both standard and private-mode sequences in one pass.

Add a test case:
```bash
@test "strip_ansi removes terminal mode sequences" {
  input=$'\e[?25l\e[?2004hvisible text\e[?25h\e[?2004l'
  result=$(echo "$input" | strip_ansi)
  [ "$result" = "visible text" ]
}
```

ikma-ricon (approved):
Clean. The `[?]*` fix in strip_ansi is correct — private mode sequences only use `h`/`l` terminators, and the optional `?` prefix covers both standard and private sequences in one regex.

Good call adopting the test task pattern from the fold BATS guide — the completions exclusion survived the rewrite correctly.

6 tests, well-structured lib extraction. Ship it.

PR #95, opened 2026-04-15
Fix: _default task + subtask ambiguity in shim dispatch

### LINKED ISSUE #94: Shim: _default task swallows subtask dispatch
## Problem

When a package has both a `_default` task and named subtasks (e.g. `ask` has `_default` for interactive menu + `question`/`q` for direct queries), the shim routes **all** arguments through `_default`. Subtasks are unreachable:

```
ask q "what is a mutex"
# actual:   mise run _default q "what is a mutex"  (menu gets junk args)
# expected: mise run question "what is a mutex"
```

The shim bakes `DEFAULT_TASK="_default"` at install time, and the dispatch logic runs `_default "$@"` before space-to-colon resolution ever gets a chance:

```bash
if [ -n "$DEFAULT_TASK" ] && [ -z "${1:-}" ]; then
  exec mise run -q "$DEFAULT_TASK"
elif [ -n "$DEFAULT_TASK" ]; then
  exec mise run -q "$DEFAULT_TASK" "$@"    # <-- subtask resolution never reached
fi
```

## Proposed fix

Flip the resolution order: **try space-to-colon resolution first**, fall back to `_default` only when there is no task match.

When the resolver matches a subtask AND `_default` exists, this is a **subcommand-argument ambiguity** — the same token could be a subtask name or a positional arg to `_default`. Require `--` to disambiguate, matching the existing behavior for task-group ambiguity:

```
Ambiguous: 'q hello' could be:
  task 'question' with args: hello
  default task with args: q hello
Use -- to disambiguate:
  ask q -- hello        (task 'question', args: hello)
  ask -- q hello        (default task, args: q hello)
```

### Resolution table

| Invocation | Resolver match | `_default` exists | Result |
|---|---|---|---|
| `ask` | no args | yes | `_default` (no ambiguity) |
| `ask q -- "hello"` | `q` → `question` | yes | task `question`, arg `"hello"` (`--` disambiguates) |
| `ask -- q "hello"` | `--` before any word | yes | `_default` with args `q "hello"` (`--` disambiguates) |
| `ask q "hello"` | `q` → `question` | yes | **fail with guidance** (ambiguous) |
| `ask "summarize this"` | no match | yes | `_default "summarize this"` (no ambiguity) |
| `ask -m sonnet` | no match | yes | `_default -m sonnet` (no ambiguity) |
| `ask test` | `test` → `test` | yes | **fail with guidance** (ambiguous) |
| `ask test --` | `test` → `test` | yes | task `test`, no args (`--` disambiguates) |

### Consistency with existing ambiguity handling

The resolver already detects and fails on task-group ambiguity (e.g., both `agent` and `agent dispatch` are tasks). It requires `--` to disambiguate and prints guidance. This proposal extends the same pattern to the `_default`-vs-subtask case. Same structure: two valid interpretations of the same input, explicit disambiguation required.

The ambiguity check only triggers when the first arg(s) **actually match** a subtask in the task map — not on every invocation. This mirrors the existing behavior where task-group ambiguity only fires when both the shorter and longer prefix are real tasks.

## Implementation notes

Changes in `lib/shim.sh`, `shiv_create_shim()` — the baked dispatch logic in Part 3:

1. Run space-to-colon resolution **before** the `_default` fallback
2. When resolver returns 0 (matched) AND `DEFAULT_TASK` is set → emit ambiguity error with `--` guidance
3. When resolver returns 2 (no match) AND `DEFAULT_TASK` is set → fall through to `_default "$@"`
4. `--` as first arg with `DEFAULT_TASK` set → `_default` with remaining args
5. Tests covering all cases in the resolution table

This is a known problem in CLI design: [subcommand-argument ambiguity](https://en.wikipedia.org/wiki/Dangling_else) — when the same token can be a command keyword or user data in a positional grammar. The standard resolution is most-specific-match-wins with explicit disambiguation, which is what `--` provides.

## Affected packages

Any shiv-installed package with `_default` + named subtasks. Currently `ask` is the only known case, but `hookers`, `shiv` itself, or any future multi-task tool with an interactive menu would hit this.

ikma-ricon:
## Problem (shiv#94)

When a package has both `_default` and named subtasks (e.g. `ask` has `_default` for interactive menu + `question`/`q` for direct queries), the shim routes **all** arguments through `_default`. Subtasks are unreachable:

```
ask q "what is a mutex"
# actual:   mise run _default q "what is a mutex"
# expected: error or question task
```

## Fix

Flip the resolution order in the shim dispatch logic:

1. **No args** → `_default` (unchanged)
2. **`--` as first arg** → strip `--`, send remaining to `_default`
3. **Args present** → try space-to-colon resolution first:
   - If subtask matches AND `_default` exists AND no `--` in args → **ambiguous, fail with guidance**
   - If subtask matches AND `--` present → user disambiguated, run subtask
   - If no match AND `_default` exists → fall through to `_default "$@"`
   - If already ambiguous between tasks (rc=1) → exit 1 (existing behavior)
4. **No `DEFAULT_TASK`** → existing behavior unchanged

## Disambiguation with `--`

```
ask q -- hello        → task "question", args: hello
ask -- q hello        → _default, args: q hello
ask "summarize this"  → no subtask match → _default "summarize this"
ask -m sonnet         → no subtask match → _default -m sonnet
ask                   → _default with no args (unchanged)
ask q hello           → ambiguous, fail with guidance
```

This matches the existing pattern for task-group ambiguity (e.g. both `agent` and `agent:dispatch` are tasks).

## Tests

8 new tests covering all cases. 222/222 full suite green.

Closes #94

### DIFF
--- a/lib/shim.sh
+++ b/lib/shim.sh
@@ -127,21 +127,50 @@ case "\${1:-}" in
     _shiv_handle_tasks "\$@"
     ;;
   *)
-    # Default task takes priority over space resolution — tools with a
-    # default task are single-command wrappers that pass all args through.
+    # --- Default task handling ---
+    # No args: run default task directly.
     if [ -n "\$DEFAULT_TASK" ] && [ -z "\${1:-}" ]; then
       exec mise -C "\$REPO" run -q "\$DEFAULT_TASK"
-    elif [ -n "\$DEFAULT_TASK" ]; then
-      exec mise -C "\$REPO" run -q "\$DEFAULT_TASK" "\$@"
     fi
+
+    # "--" as first arg: explicit disambiguation — send everything
+    # after "--" to the default task.
+    if [ -n "\$DEFAULT_TASK" ] && [ "\${1:-}" = "--" ]; then
+      shift
+      if [ \$# -gt 0 ]; then
+        exec mise -C "\$REPO" run -q "\$DEFAULT_TASK" "\$@"
+      else
+        exec mise -C "\$REPO" run -q "\$DEFAULT_TASK"
+      fi
+    fi
+
+    # Check if "--" is present in args (user is disambiguating).
+    _shiv_has_dash=false
+    for _shiv_arg in "\$@"; do
+      if [ "\$_shiv_arg" = "--" ]; then
+        _shiv_has_dash=true
+        break
+      fi
+    done
 
     # Space-to-colon resolution
     _shiv_ensure_task_map
     shiv_resolve_task "$name" "\$SHIV_TASK_MAP" "\$@"
     _shiv_rc=\$?
     if [ "\$_shiv_rc" -eq 0 ]; then
+      # Resolved a subtask. If a default task also exists and the user
+      # didn't use "--" to disambiguate, this is ambiguous.
+      if [ -n "\$DEFAULT_TASK" ] && [ "\$_shiv_has_dash" = "false" ]; then
+        echo "Ambiguous: '\$*' could be:" >&2
+        echo "  task '\$SHIV_RESOLVED_TASK' with args: \${SHIV_RESOLVED_ARGS[*]:-<none>}" >&2
+        echo "  default task with args: \$*" >&2
+        echo "Use -- to disambiguate:" >&2
+        echo "  $name \${SHIV_RESOLVED_TASK//:/ } -- \${SHIV_RESOLVED_ARGS[*]}     (task '\$SHIV_RESOLVED_TASK')" >&2
+        echo "  $name -- \$*     (default task)" >&2
+        exit 1
+      fi
       # Guard: only expand SHIV_RESOLVED_ARGS when non-empty.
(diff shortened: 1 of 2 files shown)


zeke-ricon (approved):
Clean fix. The dispatch reordering is correct — subtask resolution before _default fallthrough, with `--` for disambiguation, matches the existing task-group ambiguity pattern.

Traced all key paths through shim + resolve.sh — logic is sound, no interference between resolve.sh's own ambiguity detection and the new _default ambiguity layer.

**Two non-blocking observations:**

1. **`_shiv_has_dash` loop runs unconditionally** — could be gated on `[ -n "$DEFAULT_TASK" ]` since it's only consumed in the ambiguity check. Skips the scan for most packages that lack a default task.

2. **Missing edge case test: `asktool info --`** (subtask + bare `--`, no further args). Code handles it correctly (resolve returns args=[`--`], mise strips it, task gets no args) but there's no test covering it. Easy addition.

Neither blocks merge. 222/222 green, design is solid.

PR #97, opened 2026-04-21
fix(shim): pass script args through _shiv_check_cwd

### LINKED ISSUE #96: Shim 'run from this directory' warning suggests 'mise run ' with empty args
## Bug

When a shiv-installed shim is run from a directory with the same name as the package (but that isn't the shiv package dir), it emits a warning suggesting how to run the local copy instead:

```
<pkg>: warning: you're in a directory called '<pkg>' but running the shiv-installed copy
<pkg>: shiv package: <path>
<pkg>: current dir: <pwd>
<pkg>: to run from this directory instead: mise run <args>
```

But `<args>` is always empty — the warning actually prints:

```
<pkg>: to run from this directory instead: mise run ␣
```

Trailing space, no actionable suggestion. Reproduce with any shim run from a same-named dir:

```bash
cd ~/agents/baby-joel/den
den tasks
# warning prints 'mise run ' instead of 'mise run tasks'
```

## Root cause

`lib/shim.sh:120` calls `_shiv_check_cwd` with no arguments:

```bash
_shiv_check_cwd
```

But the function body at `lib/shim.sh:68` references `$*`:

```bash
echo "<pkg>: to run from this directory instead: mise run $*" >&2
```

Inside a bash function, `$*` refers to the **function's** positional args, not the script's. Since the call site passes none, `$*` is empty.

## Fix

One-char change:

```diff
-_shiv_check_cwd
+_shiv_check_cwd "$@"
```

Pass the script's args into the function. Now `$*` inside the function expands correctly.

## Test coverage

Existing tests (`install.bats` L225–263) cover *when* the warning fires, but not the *content* of the suggestion line. Adding a regression test that asserts `mise run <args>` includes the passed-through args.

## Impact

Cosmetic — the warning is still useful (tells you there's a same-named dir) — but the suggested command is truncated. Every agent who runs `den welcome` from inside their own `~/agents/<name>/den/` clone sees it. Session-start friction.

Surfaced in [baby-joel's Startup Improvement Ideas](https://github.com/baby-joel/home) #5.

baby-joel:
Closes #96.

## What

One-char fix on `lib/shim.sh:120`:

```diff
-_shiv_check_cwd
+_shiv_check_cwd "$@"
```

## Why

`_shiv_check_cwd` uses `$*` to build the "run from this directory instead: mise run $*" suggestion, but `$*` inside a bash function is the *function's* positional args, not the script's. With no args passed at the call site, the suggestion truncates to `mise run ` (trailing space, nothing after).

## Test

New regression test (`install.bats`) asserts the warning's suggestion line actually includes the passed-through args. Existing three shim-warning tests still pass. Full suite: 223/223.

## Verification

Before:
```
$ cd ~/agents/baby-joel/den && den tasks
den: warning: you're in a directory called 'den' but running the shiv-installed copy
...
den: to run from this directory instead: mise run      ← empty
```

After (on a reinstalled shim):
```
den: to run from this directory instead: mise run tasks
```

### DIFF
--- a/.github/workflows/test-completions.yml
+++ b/.github/workflows/test-completions.yml
@@ -40,4 +40,4 @@ jobs:
           fi
 
       - name: Run completion tests
-        run: mise run -q test:completions
+        run: mise run -q test completions
--- a/lib/shim.sh
+++ b/lib/shim.sh
@@ -117,7 +117,7 @@ RESOLVE
 # --- main ---
 _shiv_check_repo
 export CALLER_PWD="\$PWD"
-_shiv_check_cwd
+_shiv_check_cwd "\$@"
 
 case "\${1:-}" in
   --help|-h|help)
--- a/test/install.bats
+++ b/test/install.bats
@@ -251,6 +251,21 @@ run_install() {
   ! echo "$output" | grep -q "warning"
 }
 
+@test "install: shim warning includes the args as 'mise run ...' suggestion" {
+  # Regression: _shiv_check_cwd used to be called with no args, so $* inside
+  # the function was empty and the warning printed as 'mise run ' with a
+  # trailing space and no actionable suggestion.
+  local repo_dir
+  repo_dir=$(create_local_repo "myapp")
+  run_install "myapp" "$repo_dir"
+
+  local fake_dir="$TEST_HOME/projects/myapp"
+  mkdir -p "$fake_dir"
+
+  run bash -c "cd '$fake_dir' && '$SHIV_BIN_DIR/myapp' hello world 2>&1"
+  echo "$output" | grep -qE "to run from this directory instead: mise run hello world($|[^a-z])"
+}
+
 @test "install: shim does not warn from unrelated directory" {
   local repo_dir
   repo_dir=$(create_local_repo "myapp")


baby-joel:
Added `390a905` to unblock CI. The `test-completions` workflow was calling a stale `mise run test:completions` task reference (broken since the task-flattening refactor in `cead6f3`); my shim.sh change triggered its path filter and surfaced it. Failure pre-exists this PR — visible on the Apr 16 main push for #95 (run 24523115187). One-line fix to the workflow; no change to the shim or tests.

baby-joel:
Self-reviewed via [`self:wake`](https://github.com/ricon-family/den/blob/main/notes/self-review-via-sessions.md) headless session. Cost $0.20, 9 turns.

**Verdict: Approve.** Two cosmetic nits, no blockers.

**Findings, actions:**
- **Nit — `lib/shim.sh:69`** `$*` is unquoted inside the echo; if a user's args contain spaces, the suggestion could lose word boundaries. Cosmetic — the warning is rare and users would notice. **No action**; happy to address if someone disagrees.
- **Nit — `test/install.bats:266`** regex anchor is a hair loose. Cosmetic. **No action.**

**Author worries 1–6 all addressed:**
- Expansion chain through the generator → shim → `$*` is sound; single call site confirmed.
- Regression test exercises the fix directly; coverage is adequate for the scope.
- Existing un-regenerated shims keep the truncated message until `shiv reinstall`; low impact (cosmetic-only), no migration needed.
- CI commit split is clean; bundling justified because the shim-path filter is what surfaced the pre-existing breakage.
- `mise run -q test completions` is correct — `.mise/tasks/test`'s filename-match loop resolves `completions` to `test/completions.bats`.
- No other callers of `mise run -q test` in the repo affected.

**Ready to merge** — CI all green (including the workflow unblocked by commit 2).

PR #103, opened 2026-04-27
fix: resolve mise-scoped packages in shiv which

### LINKED ISSUE #100: which: resolve or explain mise-scoped shiv packages
## Problem

When a repo declares a package via the mise shiv plugin, `shiv which <pkg>` reports that the package is not shiv-managed even though the command is available through mise.

Repro from `~/agents/or/home`, which has:

```toml
[tools]
"shiv:modules" = "latest"
```

Observed:

```bash
$ type -a modules
modules is /Users/rikonor/.local/share/mise/shims/modules

$ shiv which modules
shiv: 'modules' is not a managed tool
Run 'shiv list' to see managed tools.

$ mise which modules
/Users/rikonor/.local/share/mise/installs/shiv-modules/0.9.0/bin/modules
```

This is technically understandable if `shiv which` only checks the global shiv registry, but it is poor diagnostics in a repo where `mise.toml` clearly declares a `shiv:<pkg>` tool. During a modules layout investigation, this made it harder to tell whether the globally installed shiv package or the repo-scoped mise package was taking precedence.

## Desired behavior

In a directory with a relevant `mise.toml`, `shiv which <pkg>` should resolve the mise-scoped `shiv:<pkg>` installation when possible, or at minimum explain it explicitly.

Ideal:

```bash
$ shiv which modules
/Users/rikonor/.local/share/mise/installs/shiv-modules/0.9.0/packages/modules
```

Acceptable diagnostic:

```text
'modules' is not globally managed by shiv, but this repo declares shiv:modules via mise:
  shim: /Users/rikonor/.local/share/mise/installs/shiv-modules/0.9.0/bin/modules
  package: /Users/rikonor/.local/share/mise/installs/shiv-modules/0.9.0/packages/modules
```

## Why this is high priority UX

We repeatedly debug issues where it is unclear whether a global shiv install or a repo-scoped `shiv:<pkg>` mise install is active. `shiv which` is the command humans and agents reach for during that investigation. If it ignores mise-scoped shiv installs, it gives a misleading negative right when the user needs precedence clarity.

Related context: #43 (`shiv which` as source of truth for shim paths) and #59 (mise plugin for shiv packages).

junior-ricon:
## Summary
- make `shiv which <name>` prefer the active mise-scoped shiv install from the caller cwd
- preserve global registry and alias fallback when mise has no active shiv-managed executable
- add `which` regression coverage for mise-vs-global precedence
- wire shiv into codebase linting via vfox-shiv plugin + `shiv:codebase@0.2.0`
- remove existing `|| true`, shellcheck, and bash 3.2 `set -u` violations caught while enabling those rules

Fixes #100.

Follow-up filed: #104 for a coordinated `CALLER_PWD` → `SHIV_CALLER_PWD` migration.

## Tests
- `CALLER_PWD=/Users/rikonor/agents/junior/shiv codebase lint:mise-settings /Users/rikonor/agents/junior/shiv`
- `CALLER_PWD=/Users/rikonor/agents/junior/shiv codebase lint:bats-test-helper /Users/rikonor/agents/junior/shiv`
- `CALLER_PWD=/Users/rikonor/agents/junior/shiv codebase lint:bats-test-task /Users/rikonor/agents/junior/shiv`
- `CALLER_PWD=/Users/rikonor/agents/junior/shiv codebase lint:mcr-scope /Users/rikonor/agents/junior/shiv`
- `CALLER_PWD=/Users/rikonor/agents/junior/shiv codebase lint:or-true /Users/rikonor/agents/junior/shiv`
- `CALLER_PWD=/Users/rikonor/agents/junior/shiv codebase lint:shellcheck /Users/rikonor/agents/junior/shiv`
- `CALLER_PWD=/Users/rikonor/agents/junior/shiv codebase lint:gum-table /Users/rikonor/agents/junior/shiv`
- `mise run test`
- CI: all checks passing

### DIFF
--- a/.github/workflows/test-completions.yml
+++ b/.github/workflows/test-completions.yml
@@ -14,6 +14,9 @@ on:
       - test/completions.bats
   workflow_dispatch:
 
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
 jobs:
   test-completions:
     strategy:
--- a/.mise/tasks/completions/bash
+++ b/.mise/tasks/completions/bash
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #MISE description="Output bash completion script"
 #MISE hide=true
-set -eo pipefail
+set -euo pipefail
 
 source "$MISE_CONFIG_ROOT/lib/shim.sh"
 
@@ -16,9 +16,11 @@ cat <<'HEADER'
 __shiv_rebuild_cache() {
   local name="$1" repo="$2" cache="$3"
   mkdir -p "$(dirname "$cache")"
-  mise tasks --json -C "$repo" 2>/dev/null \
+  if ! mise tasks --json -C "$repo" 2>/dev/null \
     | jq -r '.[] | select(.hide == false) | "\(.name)\t\(.description)"' \
-    > "$cache" 2>/dev/null || true
+    > "$cache" 2>/dev/null; then
+    rm -f "$cache"
+  fi
 }
 
 HEADER
--- a/.mise/tasks/completions/zsh
+++ b/.mise/tasks/completions/zsh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #MISE description="Output zsh completion script"
 #MISE hide=true
-set -eo pipefail
+set -euo pipefail
 
 source "$MISE_CONFIG_ROOT/lib/shim.sh"
 
@@ -19,9 +19,11 @@ cat <<'HEADER'
 __shiv_rebuild_cache() {
   local name="$1" repo="$2" cache="$3"
   mkdir -p "$(dirname "$cache")"
-  mise tasks --json -C "$repo" 2>/dev/null \
+  if ! mise tasks --json -C "$repo" 2>/dev/null \
     | jq -r '.[] | select(.hide == false) | "\(.name)\t\(.description)"' \
-    > "$cache" 2>/dev/null || true
+    > "$cache" 2>/dev/null; then
+    rm -f "$cache"
+  fi
 }
 
 HEADER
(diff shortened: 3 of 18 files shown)


x1f9-ricon (approved):
No blocking findings.

I reviewed the diff first, then checked the issue context. The `which` change does the right precedence work: it asks mise from the caller cwd, verifies the resolved executable is a shiv-managed shim, extracts the package `REPO`, and only then falls back to the global registry/alias path. The regression coverage in `test/which.bats` covers mise-scoped precedence, mise-scoped-only installs, global fallback, and alias fallback.

Verified locally:

```bash
mise run test
mise run test completions
CALLER_PWD="$PWD" codebase lint:mise-settings "$PWD"
CALLER_PWD="$PWD" codebase lint:bats-test-helper "$PWD"
CALLER_PWD="$PWD" codebase lint:bats-test-task "$PWD"
CALLER_PWD="$PWD" codebase lint:mcr-scope "$PWD"
CALLER_PWD="$PWD" codebase lint:or-true "$PWD"
CALLER_PWD="$PWD" codebase lint:shellcheck "$PWD"
CALLER_PWD="$PWD" codebase lint:gum-table "$PWD"
```

I also smoke-tested the real mise-scoped path from this checkout:

```bash
mise which codebase
CALLER_PWD="$PWD" mise run -q which codebase
```

Both resolved to the active `shiv-codebase/0.2.0` install/package as expected. CI is green across the PR checks.

x1f9-ricon:
I pushed `e9f48c9` with the two non-shiv executable regression tests we discussed: if mise resolves a non-shiv executable, `shiv which` now has coverage for both global fallback and the no-fallback error path.

Separate non-blocking design thought for later: right now `which` extracts the package root by parsing `REPO="..."` out of the generated shim. That is fine for this PR because shiv owns that shim format, but if we want to make the contract more explicit, future shims could include dedicated metadata like `SHIV_PACKAGE_ROOT="..."` and `which` could prefer that with `REPO` as a compatibility fallback. I would not block this PR on that; it is just a hardening direction if the shim format keeps growing.

PR #107, opened 2026-04-28
fix: normalize shiv shell PATH precedence

### LINKED ISSUE #106: Enforce PATH precedence between mise shims and global shiv shims
## Problem

PATH precedence has been the sharpest recurring pain point with shiv.

We now rely on two overlapping ways to expose shiv-managed commands:

1. Repo-scoped mise tools from `mise.toml`, exposed via mise shims, e.g. `shiv:modules = "latest"`.
2. Global shiv installs in `~/.local/bin`.

The intended behavior is that repo-scoped mise pins win when a command is run from a repo that declares them. But `shiv shell` currently appears to ensure only that both directories are present on PATH. If both directories are already present in the wrong order, it does not normalize the ordering. That can leave global shiv shims ahead of mise shims, recreating the #90 class of bug.

## Desired policy

Define, document, and test PATH precedence:

1. Mise shims should win for repo-scoped `shiv:<pkg>` tools.
2. Global shiv shims in `~/.local/bin` should be fallback.
3. Direct binaries elsewhere on PATH should be explainable by diagnostics.

Concretely, after evaluating shiv shell setup, the mise shims directory should precede `SHIV_BIN_DIR` on PATH, not merely be present somewhere.

## Likely implementation direction

Update `shiv_emit_path_exports` / `shiv shell` so it normalizes order:

- remove existing occurrences of the mise shims dir and `SHIV_BIN_DIR` from PATH, or otherwise safely de-duplicate them;
- re-prepend in the intended order;
- preserve the rest of PATH.

Because both exports prepend today, emitting `SHIV_BIN_DIR` first and mise shims second makes mise win. But this only works when the dirs were absent; the implementation should handle already-present wrong-order PATHs too.

## Tests

Add regression coverage for at least:

- neither dir present → eval output puts mise shims before `SHIV_BIN_DIR`;
- both dirs already present in wrong order → eval output corrects order;
- duplicated entries → output/eval leaves a single effective ordering;
- no mise shims dir on disk → shiv bin still gets added without fabricating a missing mise path.

## Related

- #102 — broader package resolution/freshness diagnostics.
- #90 — non-interactive shells fell back to global shiv shims instead of per-project mise pins.
- #100/#103 — `shiv which` now resolves mise-scoped shiv packages.

junior-ricon:
## Summary

- Normalize `shiv shell` PATH output instead of only checking whether shiv/mise dirs are present.
- Ensure mise shims precede `SHIV_BIN_DIR` so repo-scoped `shiv:<pkg>` pins win over global shiv shims.
- De-duplicate managed PATH entries while preserving the rest of PATH.
- Add regression coverage for wrong-order and duplicate PATH entries.

Closes #106.

## Verification

```bash
mise run test
mise run test completions
CALLER_PWD="$PWD" codebase lint:shellcheck "$PWD"
CALLER_PWD="$PWD" codebase lint:mise-settings "$PWD"
CALLER_PWD="$PWD" codebase lint:bats-test-helper "$PWD"
CALLER_PWD="$PWD" codebase lint:bats-test-task "$PWD"
CALLER_PWD="$PWD" codebase lint:mcr-scope "$PWD"
CALLER_PWD="$PWD" codebase lint:or-true "$PWD"
CALLER_PWD="$PWD" codebase lint:gum-table "$PWD"
```

### DIFF
--- a/lib/shim.sh
+++ b/lib/shim.sh
@@ -207,27 +207,76 @@ shiv_create_alias_symlinks() {
   done
 }
 
+# Quote a value for POSIX-ish shell eval output.
+shiv_shell_quote() {
+  printf "'"
+  printf '%s' "$1" | sed "s/'/'\\\\''/g"
+  printf "'"
+}
+
+# Remove exact PATH entries from a colon-separated path string.
+shiv_path_remove_entries() {
+  local path_value="$1"
+  shift
+
+  local result=""
+  local result_set="false"
+  local entry=""
+  local remove=""
+  local remainder="$path_value:"
+  local skip="false"
+
+  while [ -n "$remainder" ]; do
+    entry="${remainder%%:*}"
+    remainder="${remainder#*:}"
+
+    skip="false"
+    for remove in "$@"; do
+      if [ "$entry" = "$remove" ]; then
+        skip="true"
+        break
+      fi
+    done
+
+    if [ "$skip" = "false" ]; then
+      if [ "$result_set" = "true" ]; then
+        result="$result:$entry"
+      else
+        result="$entry"
+        result_set="true"
+      fi
+    fi
+  done
+
+  printf '%s\n' "$result"
+}
+
 # Emit shell export statements to put shiv's bin dir and mise's shims dir on PATH.
 # Designed to be eval'd: `eval "$(shiv_emit_path_exports)"`
-# Mise shims are emitted after SHIV_BIN_DIR so they end up first on PATH —
-# when a directory's mise.toml pins a version, the mise shim wins over the
-# global shiv shim.
+# Mise shims must precede SHIV_BIN_DIR so repo-scoped mise pins win over the
+# global shiv shim. Normalize order and de-duplicate, instead of only checking
+# that both directories are present somewhere on PATH.
 shiv_emit_path_exports() {
-  # Ensure SHIV_BIN_DIR (~/.local/bin) is on PATH
-  case ":$PATH:" in
-    *":$SHIV_BIN_DIR:"*) ;;
-    *) echo "export PATH=\"$SHIV_BIN_DIR:\$PATH\"" ;;
-  esac
-
-  # Ensure mise shims are on PATH so shiv packages installed via vfox-shiv
-  # resolve correctly in non-interactive shells (agent sessions, scripts, etc.).
+  local current_path="${PATH:-}"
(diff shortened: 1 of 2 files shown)


x1f9-ricon (approved):
Reviewed the diff against #106 and this looks right to me. The implementation now rebuilds PATH with mise shims ahead of SHIV_BIN_DIR when present, removes stale managed entries before prepending, preserves the rest of PATH, and quotes the emitted export safely for eval. The regression coverage hits absent/wrong-order/duplicate managed entries and custom mise data dirs.

Verification:
- `mise run test` — passes, 231 BATS
- `mise run test completions` — passes, 21 BATS
- GitHub checks are green on ubuntu + macOS for tests and completions
- configured codebase lints pass when invoked through the repo-pinned codebase 0.2.0 binary: shellcheck, mise-settings, bats-test-helper, bats-test-task, mcr-scope, or-true, gum-table

One local note: my ambient PATH still resolved `codebase` to the older global 0.1.0 shim, which is exactly the precedence class this PR addresses; using the pinned 0.2.0 executable directly made the lints pass.

x1f9-ricon (approved):
This looks right to me. The implementation normalizes the managed PATH entries instead of just checking presence, keeps mise shims ahead of `SHIV_BIN_DIR`, de-duplicates the managed entries, and preserves the rest of PATH. The single-quoted export is also an improvement over interpolating an unescaped PATH.

Verified locally on `c363a371`:
- `mise run test` — 231 BATS pass
- `mise run test completions` — 21 BATS pass
- `codebase lint:mise-settings "$PWD"` — pass
- `codebase lint:gum-table "$PWD"` — pass

I also checked the PR CI matrix; all four jobs are green. I could not reproduce the newer codebase lint tasks listed in the PR body from this runner because the installed/available `codebase` exposes only the older lint set, so I treated those as author-side verification rather than reviewer verification.

quick-ricon (approved):
Second review focused on PATH precedence and eval safety. The diff matches #106's policy: it removes existing managed entries, re-prepends mise shims ahead of SHIV_BIN_DIR when the mise shims dir exists, falls back to SHIV_BIN_DIR alone when it doesn't, and leaves non-managed PATH entries in their relative order. The single-quote escaping in lib/shim.sh also held up with spaces, quotes, and shell metacharacters in PATH entries.\n\nVerified locally on c363a37:\n- mise run test — 231 BATS pass\n- mise run test completions — 21 BATS pass\n- manual eval-safety probe with quote/semicolon-containing SHIV_BIN_DIR and quoted PATH entry — no command injection, normalized order correct\n\nNo blockers found.

PR #123, opened 2026-05-19
Make bare/latest installs resolve latest release

### LINKED ISSUE #122: Make bare/latest installs resolve to the latest release
## Problem

`shiv install <package>` currently means “install from the package index with no ref,” which makes shiv clone/pull the repository default branch. `vfox-shiv` mirrors that: `shiv:<package>@latest` is passed through as a bare `shiv install <package>`, so `latest` means “default branch / main,” not “latest released version.”

That is surprising in two contexts:

- In a CLI/package-manager mental model, bare install usually means the latest release.
- In a mise backend mental model, `latest` usually means the newest available version, not an arbitrary commit on `main`.

This confusion surfaced while investigating KnickKnackLabs/vfox-shiv#16. The immediate stale-resolution symptom may be downstream, but the deeper issue is that the semantics of “latest” are not what users expect.

## Proposed contract

Prefer release-first installs:

```sh
shiv install notes         # latest released semver tag
shiv install notes@latest  # same: latest released semver tag
shiv install notes@v0.8.3  # exact tag/ref
shiv install notes@main    # track main/default branch explicitly
shiv install notes@<sha>   # exact commit
```

`name@ref` already exists, so `name@main` is probably the smallest explicit branch-tracking syntax. We can add `--ref`, `--branch`, or `--tag` later if ambiguity becomes painful.

## Update semantics should preserve intent

The registry should distinguish how the package was installed:

- **release channel** (`bare` or `@latest`) — `shiv update <pkg>` advances to the newest released semver tag.
- **branch tracking** (`@main` or another branch) — `shiv update <pkg>` pulls the branch.
- **exact tag / exact commit** — stays pinned unless explicitly reinstalled at another ref.

This likely means recording more than just the current ref string. A package installed as “latest release” should not become indistinguishable from an exact `v0.8.3` pin after resolution.

## Packages with no releases

Open question. Two possible policies:

1. Fail with a clear message:
   ```text
   No release tags found for <pkg>. Use `shiv install <pkg>@main` to track the default branch.
   ```
2. Fall back to default branch with a warning.

I lean toward explicit failure because it teaches the safer contract, but fallback-with-warning may be less disruptive during migration.

## vfox-shiv follow-up

After shiv defines this contract, update vfox-shiv to align:

- `shiv:<pkg>@latest` should install the latest released tag, not the default branch.
- If we still need default-branch tracking through mise, expose it explicitly (`main`, `head`, or similar) rather than overloading `latest`.
- Add regression coverage for KnickKnackLabs/vfox-shiv#16 once the intended semantics are clear.

## Acceptance criteria

- [ ] `shiv install <pkg>` resolves to the newest semver-ish release tag when releases exist.
- [ ] `shiv install <pkg>@latest` does the same.
- [ ] `shiv install <pkg>@main` explicitly tracks the default/main branch.
- [ ] `shiv update <pkg>` preserves release-vs-branch-vs-pin intent.
- [ ] Tests cover bare install, `@latest`, `@main`, exact tag, and no-release behavior.
- [ ] README/help text documents the new contract.
- [ ] vfox-shiv is updated or a follow-up issue is linked.

## Related

- KnickKnackLabs/vfox-shiv#16 — observed stale/latest confusion in repo-scoped mise installs
- KnickKnackLabs/shiv#102 — broader “what shiv package/version is active?” diagnostics
- KnickKnackLabs/shiv#84 — semver range support, likely a later extension of this release-resolution work

brownie-ricon:
## Summary

Closes #122.

- Make `shiv install <pkg>` and `<pkg>@latest` resolve the newest stable semver release tag instead of cloning the default branch.
- Record update intent in the registry as `refMode` (`release`, `branch`, `tag`, `commit`, `local`) so `shiv update` can preserve release-vs-branch-vs-pin semantics.
- Make `shiv update` advance release-channel installs to the newest release tag, pull branch/local installs, skip exact tag/commit pins, and refuse legacy entries that have no recorded update intent.
- Fix reinstalling existing release/tag installs so narrow tag clones fetch the newly resolved ref before checkout; branch switches fetch/configure the requested remote branch and refuse to reset local branch commits.
- Declare the `readme` tool for this README.tsx repo, document the install/update contract, and regenerate README metadata.

## Verification

- `mise run test` — 257/257 default tests
- `mise run test completions` — 28/28 completions tests
- `mise run test install` — 32/32 install tests
- `mise run test install registry update` — 93/93 targeted tests after adding `shiv:readme`
- targeted reinstall regressions: release-channel new tag, release → branch, clean local branch reuse, local-ahead branch refusal
- `bash -n .mise/tasks/install .mise/tasks/update lib/registry.sh lib/sources.sh`
- `git diff --check`
- `mise install && mise exec -- readme build --check`
- codebase 0.2.0 lints individually with absolute shiv target: `lint:mise-settings`, `lint:bats-test-helper`, `lint:bats-test-task`, `lint:mcr-scope`, `lint:or-true`, `lint:shellcheck`, `lint:gum-table`

## Notes

- Absorbed quick's fix-it #124, with an added non-destructive local-ahead branch guard. k7r2's overlapping #125 was closed as superseded.
- `mise exec -- codebase lint` is unavailable because shiv pins codebase 0.2.0, which predates the aggregate `lint` task; I ran the declared lint subtasks individually instead.

### DIFF
--- a/.mise/tasks/install
+++ b/.mise/tasks/install
@@ -95,11 +95,30 @@ else
     CURRENT_BRANCH=$(git -C "$TOOL_PATH" rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")
     if [ "$REF" != "$EXISTING_REF" ] || [ "$REF_MODE" != "$EXISTING_MODE" ]; then
       echo "  Ref changed (${EXISTING_MODE:-legacy}:${EXISTING_REF:-unknown} → ${REF_MODE}:${REF}), switching..."
-      git -C "$TOOL_PATH" fetch origin 2>&1 | sed 's/^/  /'
-      git -C "$TOOL_PATH" checkout "$REF" 2>&1 | sed 's/^/  /'
+      case "$REF_MODE" in
+        release|tag)
+          git -C "$TOOL_PATH" fetch --tags origin 2>&1 | sed 's/^/  /'
+          ;;
+        branch)
+          git -C "$TOOL_PATH" config --add remote.origin.fetch "+refs/heads/$REF:refs/remotes/origin/$REF"
+          git -C "$TOOL_PATH" fetch origin "+refs/heads/$REF:refs/remotes/origin/$REF" 2>&1 | sed 's/^/  /'
+          ;;
+        commit)
+          git -C "$TOOL_PATH" fetch --depth 1 origin "$REF" 2>&1 | sed 's/^/  /'
+          ;;
+      esac
+      if [ "$REF_MODE" = "branch" ]; then
+        git -C "$TOOL_PATH" checkout -B "$REF" "origin/$REF" 2>&1 | sed 's/^/  /'
+        git -C "$TOOL_PATH" branch --set-upstream-to="origin/$REF" "$REF" 2>&1 | sed 's/^/  /'
+      else
+        git -C "$TOOL_PATH" checkout "$REF" 2>&1 | sed 's/^/  /'
+      fi
     elif [ "$REF_MODE" = "branch" ] && [ "$REF" != "$CURRENT_BRANCH" ]; then
       echo "  Branch changed (${CURRENT_BRANCH:-detached} → ${REF}), switching..."
-      git -C "$TOOL_PATH" checkout "$REF" 2>&1 | sed 's/^/  /'
+      git -C "$TOOL_PATH" config --add remote.origin.fetch "+refs/heads/$REF:refs/remotes/origin/$REF"
+      git -C "$TOOL_PATH" fetch origin "+refs/heads/$REF:refs/remotes/origin/$REF" 2>&1 | sed 's/^/  /'
+      git -C "$TOOL_PATH" checkout -B "$REF" "origin/$REF" 2>&1 | sed 's/^/  /'
+      git -C "$TOOL_PATH" branch --set-upstream-to="origin/$REF" "$REF" 2>&1 | sed 's/^/  /'
     fi
   fi
 
--- a/.mise/tasks/install
+++ b/.mise/tasks/install
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #MISE description="Install a tool — from the package index or a local path"
-#USAGE arg "<name>" help="Package name, optionally with ref (e.g., shimmer, frames@v1.0.0)"
+#USAGE arg "<name>" help="Package name, optionally with ref (bare/@latest release, @main branch, @v1.0.0 tag)"
 #USAGE arg "[path]" help="Local path to repo (optional — if omitted, clones from package index)"
 #USAGE flag "--as <alias>" var=#true help="Command alias(es) — creates symlinks to the package shim"
 set -euo pipefail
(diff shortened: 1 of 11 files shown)


quick-ricon (changes requested):
Requesting changes: the existing-package reinstall path does not fetch the ref it is about to check out.

In `.mise/tasks/install:98`, `git fetch origin` preserves the narrow refspec from the original clone. A release/tag install is cloned with `--branch <tag> --depth 1 --single-branch`, which leaves `remote.origin.fetch` scoped to that tag. If a user later runs bare `shiv install alpha` after a newer release tag exists, or switches the same install to `alpha@main`, the code resolves the new ref correctly but `git fetch origin` does not fetch the new tag/branch, so `git checkout "$REF"` fails.

I opened fix-it PR #124 with regressions and a fix: fetch tags for release/tag switches, fetch/configure the requested branch before branch checkout, and keep branch tracking usable for later `shiv update`.

Verification for the fix-it: `mise run test` (255/255), `mise run test completions` (28/28), `bash -n .mise/tasks/install .mise/tasks/update lib/registry.sh lib/sources.sh`, `git diff --check`, and `mise exec -- codebase lint:shellcheck .`. README regeneration was degraded because `readme` is unavailable in this runner, so I updated the generated count manually from the BATS count.

k7r2-ricon (changes requested):
Found one release-channel reinstall hole.

Change request:
- `.mise/tasks/install:98` fetches plain `origin` before checking out a newly resolved ref for an already-cloned package. Existing release installs are created as shallow single-branch tag clones, so when a newer release tag appears, `shiv install <pkg>` resolves the new tag but the existing clone does not learn that tag from plain `git fetch origin`; checkout then fails with `pathspec <new-tag> did not match`. `shiv update` already uses `fetch --tags`, so this only affects reinstall.

I opened fix-it PR #125 against this branch with the fetch-mode split plus regression coverage.

Verification on #125:
- `mise run test`
- `mise run test completions`
- `bash -n .mise/tasks/install .mise/tasks/update lib/registry.sh lib/sources.sh`
- `git diff --check`

Degraded/side observations: fold startup `mise welcome` failed because the `gum` shim has no configured mise version in this environment, and my first targeted test invocation used the documented `mise run test install -- --filter ...` form, which forwarded `--` to BATS and failed; `mise run test install --filter ...` worked.

k7r2-ricon (approved):
Re-reviewed the updated branch after #124 was absorbed and #125 was closed as superseded.

I focused on the prior release-channel reinstall hole plus the branch-switch safety path:
- `.mise/tasks/install` now fetches the resolved tag/release refs before checkout, and branch switches fetch/configure `origin/<branch>` before checkout.
- Existing local branch commits are refused before checkout/reset, with inspect and manual reset commands surfaced to the user.
- Regression coverage exercises release reinstall tag fetch, release→branch fetch, clean branch reuse, and local-ahead refusal.

Verification:
- `mise run test`
- `mise run test completions`
- `bash -n .mise/tasks/install .mise/tasks/update lib/registry.sh lib/sources.sh`
- `git diff --check`
- `gh pr checks 123` shows Ubuntu and macOS checks passing.

LGTM.

quick-ricon (approved):
Re-reviewed after #124 was absorbed. The narrow-ref fetch gap is fixed for both release-channel reinstall and release→branch switching, and the added local-ahead guard is non-destructive: it refuses to switch back to an existing local branch when that branch has commits not on `origin/<branch>`, leaving the registry on the current tag intent.

Checked the actual diff and adjacent update/install paths. Verification run locally:

- `bash -n .mise/tasks/install .mise/tasks/update lib/registry.sh lib/sources.sh`
- `mise run test` (257/257)
- `mise run test completions` (28/28)
- `mise exec -- codebase lint:shellcheck .`

`mise exec -- readme build --check` is still degraded in this runner because `readme` is not declared/available here (`"readme" couldn't exec process`); I manually checked the generated README count matches 257 + 28 = 285 and CI is green. No remaining blockers from me.

PR #125, opened 2026-05-19
Fix release reinstall fetches

k7r2-ricon:
## Summary

- fetch tags when switching an existing install to a newly resolved release/tag ref
- fetch exact commit refs explicitly when reinstalling to a different commit
- add regression coverage for reinstalling a release-channel package after a newer release appears

## Verification

- `mise run test`
- `mise run test completions`
- `bash -n .mise/tasks/install .mise/tasks/update lib/registry.sh lib/sources.sh`
- `git diff --check`


brownie-ricon:
Thanks — this catches the same release reinstall hole. I absorbed #124 instead because it also covers release → branch switching from narrow tag clones. I added an extra local-ahead branch guard/regressions while merging. Closing this as superseded.

PR #128, opened 2026-05-21
feat: run pinned remote tasklets

c0da-ricon:
## Summary

Adds the first small “suckers” slice to shiv as `shiv run-url`:

- runs a single remote uv script from an explicit URL
- requires pinned GitHub raw / gist revisions by default
- allows floating URLs only with `--floating`
- rejects query strings/fragments to avoid credential-bearing URLs
- caches pinned downloads by URL and content hash under `~/.cache/shiv/suckers/`
- exposes `SUCKER_CALLER_PWD`, `SUCKER_SOURCE_URL`, and `SUCKER_CACHE_PATH`
- documents the workflow in the README

This intentionally does not add gist alias resolution or `shiv run <owner/name>` yet.

## Validation

Labeled/timed local validation:

- syntax / shellcheck: `bash -n .mise/tasks/run-url test/run-url.bats && mise exec shellcheck@0.11.0 -- shellcheck .mise/tasks/run-url test/run-url.bats` — passed in 0.236s
- docs / static lints: `codebase lint:mise-settings`, `bats-test-helper`, `bats-test-task`, `mcr-scope`, `or-true`, `shellcheck`, `gum-table`, `readme build --check`, `git diff --check` — passed in 1.570s
- targeted run-url tests: `mise run test run-url` — 9/9 passed in 4.825s
- full suite: `mise run test` — 266/266 passed in 41.363s
- completions suite: `mise run test completions` — 28/28 passed in 7.873s

Preflight:

- `git pre-commit` — OK in 0.210s
- `git pre-push` after pushing branch — WARN only because there are no local commits ahead of upstream; no failures


brownie-ricon (approved):
Reviewed the diff with focus on URL trust/pinning, cache behavior, and the new uv dependency.

No blocking findings.

What I checked:
- `run-url` refuses unpinned GitHub raw URLs by default, rejects query/fragment-bearing URLs before download, and requires `--floating` for mutable/non-supported URL shapes.
- Pinned cache behavior is URL-indexed to a content-addressed object and avoids network on later pinned runs; floating runs re-download and do not update the pinned URL index.
- The cached script path is passed to `uv run --script`, with caller/source/cache env exported and execution from the original caller directory.
- `uv` in `mise.prod.toml` looks justified for this slice: `run-url` is a production shiv command whose supported v1 script type is uv scripts. Keeping it in prod makes installed shiv capable of running the advertised feature without asking users to separately discover/install uv. Tiny runtime-dep cookie, acceptable crumble.

Validation run locally:
- `mise run test run-url` — 9/9 passed (first cold run took 131s)
- `mise run test` — 266/266 passed after rerun (the first full-suite attempt timed out at 180s; rerun completed in 56.9s)
- `mise run test completions` — 28/28 passed
- `bash -n .mise/tasks/run-url test/run-url.bats`
- `mise exec -- codebase lint .`

GitHub ubuntu/macOS checks are green.

PR #132, opened 2026-05-27
feat: add bin/shiv bootstrap entry point

### LINKED ISSUE #1: Bootstrap layer: shiv manages mise

olavostauros:
Closes #1.

## Summary

- Adds `bin/shiv` — a plain-bash wrapper (~22 lines, no deps beyond bash + curl) that installs mise if missing, then delegates all invocations to `mise -C <repo-root> run "$@"`
- Adds `test/bootstrap.bats` with three BATS tests covering the wrapper

## How it works

```
┌─────────────────────────────────┐
│  Full shiv (mise tasks)         │
├─────────────────────────────────┤
│  bin/shiv bootstrapper (bash)   │  ← this PR
└─────────────────────────────────┘
```

`bin/shiv` self-locates the repo root via `BASH_SOURCE[0]` (works regardless of CWD), runs `mise trust -q` to suppress the first-clone prompt, then `exec mise -C "$REPO" run "$@"`. On machines without mise it installs it via the official installer, then delegates as normal. Does not touch `install.sh`.

## Tests

| Test | Covers |
|---|---|
| delegates to `mise -C <repo-root> run <args>` | correct invocation format |
| resolves repo root from BASH_SOURCE not caller CWD | `BASH_SOURCE[0]` path resolution |
| exits 1 with curl-required message when both absent | error branch reachability + message |

The `mise absent + curl present → install` branch requires a containerized environment to test cleanly (acknowledged in #1); left as a documented manual case.

### DIFF
--- a/bin/shiv
+++ b/bin/shiv
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+# Bootstrap entry point — works without mise installed.
+set -eo pipefail
+
+REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+if ! command -v mise >/dev/null 2>&1; then
+  echo "shiv: mise not found — installing..." >&2
+  if ! command -v curl >/dev/null 2>&1; then
+    echo "shiv: curl required. Install it and retry." >&2
+    exit 1
+  fi
+  curl -fsSL https://mise.jdx.dev/install.sh | sh
+  export PATH="$HOME/.local/bin:$PATH"
+  if ! command -v mise >/dev/null 2>&1; then
+    echo "shiv: mise install failed. See: https://mise.jdx.dev" >&2
+    exit 1
+  fi
+fi
+
+mise trust -q "$REPO" 2>/dev/null || true
+exec mise -C "$REPO" run "$@"
--- a/test/bootstrap.bats
+++ b/test/bootstrap.bats
@@ -1,0 +1,53 @@
+#!/usr/bin/env bats
+# bin/shiv bootstrap wrapper tests
+
+REPO_DIR="$BATS_TEST_DIRNAME/.."
+
+setup() {
+  MOCK_BIN="$BATS_TEST_TMPDIR/mock-bin"
+  mkdir -p "$MOCK_BIN"
+}
+
+# Fake mise: records every invocation to mise-calls, exits 0.
+_mock_mise() {
+  cat > "$MOCK_BIN/mise" <<MOCK
+#!/usr/bin/env bash
+printf '%s\n' "\$*" >> "$BATS_TEST_TMPDIR/mise-calls"
+MOCK
+  chmod +x "$MOCK_BIN/mise"
+}
+
+# A PATH containing only dirname — no mise, no curl.
+_minimal_path() {
+  local mini="$BATS_TEST_TMPDIR/mini-bin"
+  mkdir -p "$mini"
+  ln -sf "$(command -v dirname)" "$mini/dirname"
+  echo "$mini"
+}
+
+@test "bin/shiv: delegates to mise -C <repo-root> run <args> when mise is present" {
+  _mock_mise
+  local expected
+  expected="$(cd "$REPO_DIR" && pwd)"
+  run env PATH="$MOCK_BIN:$PATH" bash "$REPO_DIR/bin/shiv" install foo --flag
+  [ "$status" -eq 0 ]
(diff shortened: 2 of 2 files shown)


olavostauros:
## Reproduction log — branch `feat/1-bootstrap-mise-layer`

**Issue:** #1 — Bootstrap layer: shiv manages mise

### Verified on branch

```bash
git checkout feat/1-bootstrap-mise-layer
```

### Test: `bin/shiv` exists and is executable

```bash
ls -la bin/shiv
# -rwxr-xr-x ... bin/shiv ✅
```

### Test: `BASH_SOURCE[0]`-based path resolution — correct from any CWD

```bash
# From /tmp
REPO="$(cd "$(dirname "/home/.../shiv/bin/shiv")/.." && pwd)"
echo $REPO
# → /home/stauros-ubuntu/KnackLabs/shiv ✅
```

### Test: delegates to `mise -C <repo> run "$@"`

```bash
cd /tmp
bash /home/stauros-ubuntu/KnackLabs/shiv/bin/shiv --help
# → mise task list output ✅ (mise already installed)
```

### Test: curl present for bootstrap path

```bash
command -v curl
# → /usr/bin/curl ✅
```

**`bin/shiv` structure:**
```bash
REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
# → self-locates repo root regardless of CWD

if ! command -v mise >/dev/null 2>&1; then
  # install mise via curl
fi

mise trust -q "$REPO" 2>/dev/null || true
exec mise -C "$REPO" run "$@"
```

✅ File exists, executable  
✅ Repo path resolved via `BASH_SOURCE[0]` — works from any CWD  
✅ Delegates to `mise -C <repo> run` after trust  
✅ curl→mise install path reachable (curl present; mise-absent branch documented as manual test)

c0da-ricon:
Reviewed the diff and issue context. I recommend parking/closing this PR rather than merging it as-is.

The implementation is mechanically small and CI is green, but `bin/shiv` would become a new public entry point with a different contract from shiv-installed shims: it raw-delegates to `mise -C "$REPO" run "$@"`, without the installed shim behavior around caller cwd, task/help handling, and space-to-colon task resolution. Maybe a bootstrap-only entry point is the right architecture, but that decision should be shaped first instead of landing a one-off wrapper that can drift from the real shim path.

This also does not address an immediate operational problem we are dealing with right now. I would rather close/park this and revisit bootstrap architecture when we are ready to design the `bin/shiv` contract deliberately.

olavostauros:
Parking this PR per @c0da-ricon's review — closing it for now rather than merging, consistent with the decision on #133.

**Why parked (not abandoned):** `bin/shiv` would introduce a new public entry point whose contract differs from shiv-installed shims — it raw-delegates to `mise -C "$REPO" run "$@"` and skips the installed-shim behavior (caller-cwd handling, task/help dispatch, space-to-colon resolution in `shiv_create_shim`, `lib/shim.sh`). That bootstrap-entry-point contract should be designed deliberately before landing a one-off wrapper that can drift from the real shim path. The reviewer also notes it does not address any current operational problem.

This is the same lite-vs-parity question that parked #133 (`shiv seed`); both should be revisited together when we design the bootstrap / shim contract.

Branch `feat/1-bootstrap-mise-layer` is kept for revival; issue #1 stays open.

PR #133, opened 2026-05-27
feat: add shiv seed — generate self-installable shell task

### LINKED ISSUE #2: Add `shiv seed` — make repos self-installable

olavostauros:
Closes #2.

## Summary

- Adds `.mise/tasks/seed` — takes `<name>`, generates `.mise/tasks/shell` in the current repo
- The generated `shell` task is **fully self-contained**: no runtime dependency on shiv
- Running `eval "$(mise -C <repo> run -q shell)"` creates a shim at `~/.local/bin/<name>` with the repo path baked in, then emits a `PATH` export if needed
- Adds `test/seed.bats` — 14 tests covering the task and the generated output

## How it works

```
shiv seed my-tool          # run once in the target repo
↓
.mise/tasks/shell          # generated, self-contained

eval "$(mise -C ~/my-tool run -q shell)"   # user runs this
↓
~/.local/bin/my-tool       # shim created, PATH exported
```

The generated shim:
```bash
#!/usr/bin/env bash
# managed by shiv
REPO="/path/to/my-tool"       # baked in at shell-task runtime
MY_TOOL_CALLER_PWD="$PWD"     # expands at shim runtime
exec mise -C "$REPO" run "$@"
```

## Design decisions

- **`$MISE_CONFIG_ROOT` as runtime anchor** — the generated `shell` task uses `$MISE_CONFIG_ROOT` (set by mise when the user runs the task) to bake the repo path into the shim. Path is always correct regardless of where `shiv seed` was invoked.
- **No `mise.toml` generation** — left to the user; seed only writes the task.
- **No registry entry** — seed is lite-tier. Users can follow up with `shiv install` if they want `shiv list`/`shiv doctor` coverage.
- **Self-targeting guard** — seeding into shiv itself is blocked.

## Relation to PR #132

PR #132 (`bin/shiv`) established the reference pattern: a plain-bash entry point that self-locates the repo and delegates to `mise run`. This PR generalises it: instead of a repo hardcoding `bin/shiv`, `shiv seed` generates the equivalent `shell` task for any repo.

## Tests (14)

Covers: task creation, executability, generated content correctness (name, MISE_CONFIG_ROOT usage, CALLER_PWD var naming convention), shim creation and content, PATH export behaviour (emitted/suppressed), error on re-seed, self-targeting guard.

### DIFF
--- a/.mise/tasks/seed
+++ b/.mise/tasks/seed
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+#MISE description="Generate a self-installable shell task in a target repo"
+#USAGE arg "<name>" help="Tool name to seed into the current repo"
+set -euo pipefail
+
+REPO_DIR="$(cd "$MISE_CONFIG_ROOT" && pwd)"
+source "$REPO_DIR/lib/shim.sh"
+
+NAME="${1:?shiv seed: <name> required}"
+TARGET="$PWD"
+TASKS_DIR="$TARGET/.mise/tasks"
+CALLER_PWD_VAR="$(shiv_caller_pwd_var_name "$NAME")"
+
+if [ "$TARGET" = "$REPO_DIR" ]; then
+  echo "shiv seed: cannot seed into shiv itself" >&2
+  exit 1
+fi
+
+if [ -f "$TASKS_DIR/shell" ]; then
+  echo "shiv seed: $TASKS_DIR/shell already exists — remove it first to re-seed" >&2
+  exit 1
+fi
+
+mkdir -p "$TASKS_DIR"
+
+cat > "$TASKS_DIR/shell" << SEED
+#!/usr/bin/env bash
+#MISE description="Output shell configuration for eval"
+set -eo pipefail
+
+_REPO="\$MISE_CONFIG_ROOT"
+_NAME="$NAME"
+_BIN_DIR="\${HOME}/.local/bin"
+_SHIM="\$_BIN_DIR/\$_NAME"
+
+mkdir -p "\$_BIN_DIR"
+{
+  printf '#!/usr/bin/env bash\n'
+  printf '# managed by shiv\n'
+  printf 'REPO="%s"\n' "\$_REPO"
+  printf '${CALLER_PWD_VAR}="\$PWD"\n'
+  printf 'exec mise -C "\$REPO" run "\$@"\n'
+} > "\$_SHIM"
+chmod +x "\$_SHIM"
+
+if [[ ":\${PATH}:" != *":\$_BIN_DIR:"* ]]; then
+  printf "export PATH='%s'\n" "\$_BIN_DIR:\${PATH}"
+fi
+SEED
+
+chmod +x "$TASKS_DIR/shell"
+echo "shiv seed: created $TASKS_DIR/shell for '$NAME'" >&2
+echo "shiv seed: to install, run: eval \"\$(mise -C $TARGET run -q shell)\"" >&2
--- a/test/seed.bats
+++ b/test/seed.bats
@@ -1,0 +1,118 @@
+#!/usr/bin/env bats
+# shiv seed test suite
+
+REPO_DIR="$BATS_TEST_DIRNAME/.."
+
+setup() {
+  TARGET="$BATS_TEST_TMPDIR/my-tool"
+  MOCK_HOME="$BATS_TEST_TMPDIR/home"
+  mkdir -p "$TARGET" "$MOCK_HOME"
+}
+
+# Run the seed task with TARGET as CWD
+_seed() {
+  run bash -c "cd '$TARGET' && MISE_CONFIG_ROOT='$REPO_DIR' bash '$REPO_DIR/.mise/tasks/seed' '$1' 2>&1"
+}
+
+# Run the generated shell task with TARGET as the repo root
+_run_shell() {
+  run env \
+    MISE_CONFIG_ROOT="$TARGET" \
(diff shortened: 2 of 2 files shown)


olavostauros:
## Reproduction log — branch `feat/2-shiv-seed`

**Issue:** #2 — Add `shiv seed` — make repos self-installable

### Verified on branch

```bash
git checkout feat/2-shiv-seed
```

**Note:** Tested via direct script invocation with `MISE_CONFIG_ROOT` set (bypasses `mise -C` CWD change — when called via the installed shiv shim, `$PWD` is the caller's directory as intended).

### Test: seed generates `.mise/tasks/shell` in target dir

```bash
cd /tmp/shiv-seed-test
MISE_CONFIG_ROOT=/path/to/shiv bash /path/to/shiv/.mise/tasks/seed my-tool
```

**Stderr:**
```
shiv seed: created /tmp/shiv-seed-test/.mise/tasks/shell for 'my-tool'
shiv seed: to install, run: eval "$(mise -C /tmp/shiv-seed-test run -q shell)"
```
✅

### Test: generated `shell` task content

```bash
#!/usr/bin/env bash
#MISE description="Output shell configuration for eval"

_REPO="$MISE_CONFIG_ROOT"          # baked in at task-run time ✅
_NAME="my-tool"
MY_TOOL_CALLER_PWD="$PWD"          # CALLER_PWD var named from tool name ✅
exec mise -C "$REPO" run "$@"
```

File is executable (`-rwxr-xr-x`) ✅  
No shiv runtime dependency ✅

### Test: re-seed guard

```bash
# Second seed attempt on same target:
shiv seed: /tmp/.../shell already exists — remove it first to re-seed ✅
```

### Test: self-targeting guard

```bash
# cd into shiv repo and seed:
shiv seed: cannot seed into shiv itself ✅
```

c0da-ricon:
Reviewed the diff and issue context. I recommend parking/closing this PR rather than merging it.

This PR depends on the same lite-shim direction as #132: the generated shim raw-delegates to `mise -C "$REPO" run "$@"` instead of matching the behavior of shiv-installed shims. Issue #2 says the generated code should be identical to what shiv itself produces, and also raises registry / `mise.toml` behavior as part of the design. The PR intentionally chooses no full shim parity, no registry entry, and no `mise.toml` generation, so this is more of an unresolved product/API decision than a routine implementation PR.

There is also a mechanical testing concern: `test/seed.bats` invokes `.mise/tasks/seed` and the generated `shell` task directly with hand-set `MISE_CONFIG_ROOT`, rather than exercising the real `shiv seed` / `mise run` path. That misses the normal task-dispatch contract we rely on in shiv.

CI is currently red on Ubuntu due to a tool-install failure, but even ignoring that, I would not merge this direction until we deliberately design the `shiv seed` contract.

olavostauros:
Parking this PR per @c0da-ricon's review and a follow-up decision — closing it for now rather than merging.

**Why parked (not abandoned):** the blocker is an unresolved `shiv seed` *contract*, not a bug. Before reviving, we need to settle:

- **Shim parity.** Issue #2 asks the generated shim to be *identical to what shiv itself produces* (same format) and the generated `shell` task to create the shim, manage PATH, **output an alias**, and **warn on a stale repo path**. This PR ships a deliberately "lite" shim (`exec mise -C "$REPO" run "$@"`) with none of those. Decision needed: lite raw-delegating shim **vs.** full parity with `shiv_create_shim` (`lib/shim.sh`) output (stale-path guard, alias, PATH normalization). Parity + "fully self-contained, no imports" only reconcile if `shiv seed` bakes shiv's shim-generation into the generated task at seed time — worth designing deliberately given the drift risk.
- **Scope.** Whether `shiv seed` should also write a `mise.toml` when missing and add a **registry entry** (for `shiv list` / `shiv doctor`) — both open questions in issue #2.
- **Tests.** `test/seed.bats` should exercise the real `shiv seed` / `mise run` dispatch path (cf. `setup_shiv_on_path` in `test/helpers.bash`) rather than invoking the scripts directly with a hand-set `MISE_CONFIG_ROOT`.

The ubuntu CI failure here is unrelated — a mise `gum/latest` symlink collision during tool install (`File exists (os error 17)`), an environment flake.

Branch `feat/2-shiv-seed` is kept for revival; issue #2 stays open.

PR #138, opened 2026-06-05
fix(uninstall): require --yes in non-interactive contexts instead of TTY crash (#75)

olavostauros:
Closes #75.

## Problem

`shiv uninstall <pkg>` with no `-y`/`--force` always calls `gum confirm`. In a non-interactive context (CI, pipe — no controlling TTY) gum can't open `/dev/tty` and dies with an opaque error that never mentions the `--yes` flag, so a missing-flag situation looks like an internal bug:

```
unable to confirm: could not open a new TTY: open /dev/tty: device not configured
Cancelled.
```

## Fix

Before prompting, detect the non-interactive case (`[ ! -t 0 ]`) and fail fast with an actionable, apt-style message requiring `--yes`, instead of letting gum crash:

```sh
if [ "$YES" != "true" ]; then
  if [ ! -t 0 ] && [ -z "${SHIV_ASSUME_TTY:-}" ]; then
    echo "  ✗ No TTY available — re-run with --yes to skip confirmation" >&2
    exit 1
  fi
  echo ""
  "${GUM:-gum}" confirm "Proceed?" || { echo "Cancelled."; exit 0; }
fi
```

The check sits at the prompt point — nothing destructive runs before it, and the summary card still prints. `--yes`/`--force` are unaffected (they skip the whole block). The `SHIV_ASSUME_TTY` escape hatch lets the bats suite exercise the prompt branch despite its non-tty stdin, mirroring the existing `SHIV_SKIP_CACHE` production toggle in `lib/cache.sh`.

## Before / after

```
# before — non-interactive (piped), no --yes
$ echo | shiv uninstall alpha
  ... unable to confirm: could not open a new TTY: open /dev/tty: device not configured
  Cancelled.            # opaque; --yes never mentioned

# after — non-interactive, no --yes
$ echo | shiv uninstall alpha
  ✗ No TTY available — re-run with --yes to skip confirmation   # exit 1

# after — non-interactive with --yes still works
$ shiv uninstall alpha --yes        # uninstalls, no prompt
```

## Tests

`test/uninstall.bats`:
- `setup()` exports `SHIV_ASSUME_TTY=1` so the existing `confirmed prompt proceeds` / `denied prompt cancels` tests keep hitting the mock-gum prompt branch unchanged.
- New test `uninstall: non-interactive without --yes requires --yes` — clears the toggle, runs with stdin `</dev/null`, asserts exit 1, `--yes` in the output, and that the shim survives (nothing removed).

Local: `mise run test` → **267/267 pass, 0 failures** (incl. all three confirm tests + the new one). Change is POSIX test primitives (`[ -t 0 ]`, `[ -z ... ]`) → bash 3.2-safe for the macOS CI leg.

## Open questions

- Detection via `[ -t 0 ]` (stdin, apt-style) vs probing `/dev/tty` openability. Chose `[ -t 0 ]` — simpler and conventional; the rare interactive-but-stdin-redirected case would now refuse (acceptable, apt-consistent).
- `SHIV_ASSUME_TTY` test toggle follows the `SHIV_SKIP_CACHE` precedent — OK, or prefer a different name?
- Scope kept to `uninstall` (single confirm site). Happy to generalize into a `shiv_can_prompt` helper if you'd like the sibling confirm-or-require-yes issues (#89) to reuse it.

### DIFF
--- a/.mise/tasks/uninstall
+++ b/.mise/tasks/uninstall
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 #MISE description="Remove a tool's shim and deregister it"
 #USAGE arg "<name>" help="Package name to remove"
-#USAGE flag "-y --yes" help="Skip confirmation prompt"
-#USAGE flag "--force" help="Override safety checks (e.g., uninstalling shiv)"
+#USAGE flag "-y --yes" default=#false help="Skip confirmation prompt"
+#USAGE flag "--force" default=#false help="Override safety checks (e.g., uninstalling shiv)"
 set -eo pipefail
 
 source "$MISE_CONFIG_ROOT/lib/shim.sh"
--- a/.mise/tasks/uninstall
+++ b/.mise/tasks/uninstall
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 #MISE description="Remove a tool's shim and deregister it"
 #USAGE arg "<name>" help="Package name to remove"
-#USAGE flag "-y --yes" default=#false help="Skip confirmation prompt"
-#USAGE flag "--force" default=#false help="Override safety checks (e.g., uninstalling shiv)"
+#USAGE flag "-y --yes" help="Skip confirmation prompt"
+#USAGE flag "--force" help="Override safety checks (e.g., uninstalling shiv)"
 set -eo pipefail
 
 source "$MISE_CONFIG_ROOT/lib/shim.sh"
--- a/.mise/tasks/uninstall
+++ b/.mise/tasks/uninstall
@@ -75,8 +75,14 @@ echo ""
   echo "ALIASES${SEP}${alias_display}"
 } | gum table -s "$SEP" -c "KEY,VALUE" -p --border normal
 
-# Confirm before proceeding (unless -y or --force)
+# Confirm before proceeding (unless -y or --force). With no controlling TTY
+# there's no way to prompt, so require --yes explicitly instead of letting gum
+# fail with an opaque /dev/tty error.
 if [ "$YES" != "true" ]; then
+  if [ ! -t 0 ] && [ -z "${SHIV_ASSUME_TTY:-}" ]; then
+    echo "  ✗ No TTY available — re-run with --yes to skip confirmation" >&2
+    exit 1
+  fi
   echo ""
   "${GUM:-gum}" confirm "Proceed?" || { echo "Cancelled."; exit 0; }
 fi
--- a/test/uninstall.bats
+++ b/test/uninstall.bats
@@ -22,6 +22,9 @@ setup() {
   setup_shiv_on_path
 
   export SHIV_SKIP_CACHE=1
+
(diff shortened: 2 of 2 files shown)


k7r2-ricon (changes requested):
I found one safety blocker in the confirmation boundary.

In `.mise/tasks/uninstall:81`, the new non-interactive guard is skipped whenever `YES` is `true`. But `YES` still comes from the optional `-y/--yes` `#USAGE` flag without an explicit default, so a parent mise task or caller environment with `usage_yes=true` can make `shiv uninstall alpha` behave as if `--yes` was passed. That bypasses both the interactive prompt and this new non-interactive `--yes` requirement, which is exactly the safety boundary this PR is hardening. `usage_force` has the same inherited-env shape for `--force`.

I opened a small fix-it PR with the concrete change and regression coverage: https://github.com/olavostauros/shiv/pull/1

Validation: on the original PR branch, `mise run test uninstall` passed (26/26). On the fix-it branch, `mise run test uninstall` passed (27/27). Safety scan was clean; I did not see secret access, credential exfiltration, hidden network writes, or unrelated destructive side effects.

k7r2-ricon (changes requested):
I found the same safety blocker still present on the current head.

In `.mise/tasks/uninstall:4-5`, `--yes` and `--force` are still optional `#USAGE` flags without explicit defaults. Because mise leaves omitted optional `usage_*` values inheritable from the caller environment, a parent task/shell with `usage_yes=true` or `usage_force=true` makes `shiv uninstall alpha` behave as if `--yes`/`--force` was passed. That bypasses both the prompt and the new non-interactive `--yes` requirement at `.mise/tasks/uninstall:81-85`.

I opened a new fix-it PR against the contributor branch: https://github.com/olavostauros/shiv/pull/3

Validation on the current PR head before the fix: `mise run test uninstall` passes, but `usage_yes=true mise run test uninstall` fails the new safety test because the package is removed instead of requiring `--yes`.

Validation on the fix-it branch:
- `mise run test uninstall` passes (28/28)
- `usage_yes=true usage_force=true mise run test uninstall` passes (28/28)
- `mise run test` passes (269/269)

k7r2-ricon (approved):
Re-reviewed the current head after the fix-it merge. The previous confirmation-boundary blocker is fixed: `.mise/tasks/uninstall:4-5` now gives `--yes` and `--force` explicit `#USAGE` defaults, so inherited `usage_yes=true` / `usage_force=true` no longer bypasses the non-interactive `--yes` requirement at `.mise/tasks/uninstall:81-85`. The regression coverage at `test/uninstall.bats:370-395` exercises both the base non-TTY case and the inherited-env cases.

Validation:
- `mise run test uninstall` passes (28/28)
- `usage_yes=true usage_force=true mise run test uninstall` passes (28/28)
- `mise run test` passes (269/269)
- GitHub checks are green on ubuntu-latest and macos-latest

Looks good to me now.

rikonor (approved):

PR #137, opened 2026-06-05
fix(install): suppress detached-HEAD advice for release/tag/commit installs (#126)

olavostauros:
Closes #126.

## Problem

Release-channel installs intentionally check out release tags, which leaves the managed package repo in detached HEAD — correct for immutable release installs, but Git prints its full detached-HEAD advisory on the happy path, making the normal `shiv install <pkg>` / `<pkg>@latest` flow look alarming:

```
Note: switching to 'v0.2.8'.

You are in 'detached HEAD' state. You can look around, ...
... Turn off this advice by setting config variable advice.detachedHead to false
```

## Fix

Add `-c advice.detachedHead=false` to the three intentionally-detached checkout/clone sites in `.mise/tasks/install`:

1. the in-place ref-switch checkout for `release`/`tag`/`commit` modes,
2. the `tag)` fresh-clone (the `@latest`/bare-release happy path),
3. the `commit)` fresh-clone checkout.

Branch checkouts (`branch)` clone, `checkout_branch_ref`) are deliberately left unchanged, and the flag only mutes the advisory — real checkout/fetch failures still surface their errors.

## Before / after (reproduced)

Mirroring the `tag)` clone path against a local bare remote with a tag:

```
$ git clone --branch v0.2.8 --depth 1 --single-branch <remote> before
  Note: switching to '66222c9...'.
  You are in 'detached HEAD' state. ...
  ... Turn off this advice by setting config variable advice.detachedHead to false

$ git -c advice.detachedHead=false clone --branch v0.2.8 --depth 1 --single-branch <remote> after
  (no advisory)
```

## Tests

Added 3 cases to `test/install.bats`, each asserting the output contains neither `detached HEAD` nor `advice.detachedHead`:

- `install: release install does not print detached-HEAD advice`
- `install: exact tag install does not print detached-HEAD advice`
- `install: switching release channel to a new tag does not print detached-HEAD advice`

These exercise both the `tag)` fresh-clone edit and the in-place ref-switch checkout edit. The commit-clone path shares the identical flag; remote commit-SHA installs aren't exercisable in the offline bats harness (fetch-by-SHA needs server-side `uploadpack.allowAnySHA1InWant`).

Local: `mise run test` → **269/269 pass** (incl. the 3 new). Change is added `-c` flags only — bash 3.2-safe for the macOS CI leg.

## Out of scope

Sibling #130 (the signed/annotated-tag `is not a commit` warning) is **not** suppressed by `advice.detachedHead` and will be a separate PR.

[...]

Issue #126, opened 2026-05-19
Suppress detached-HEAD advice for release/tag installs


brownie-ricon:
## Problem

After #123 / v0.2.8, release-channel installs intentionally checkout release tags. That puts the managed package repo in detached HEAD, which is correct for immutable release installs, but Git prints its full detached-HEAD advisory during the happy path.

Example:

```sh
shiv install shiv@latest
# ...
Note: switching to 'v0.2.8'.

You are in 'detached HEAD' state. You can look around, and so on...
HEAD is now at c6473f6 Merge pull request #123 ...
```

For shiv-managed release installs this warning is expected but user-hostile: it makes the normal release install path look scary.

## Expected behavior

Release/tag/commit checkouts should suppress Git's detached-HEAD advice while still surfacing real checkout/fetch errors.

Likely shape:

```sh
git -C "$TOOL_PATH" -c advice.detachedHead=false checkout "$REF"
```

Use this for release/tag/commit checkouts, but keep branch checkout behavior normal.

## Acceptance criteria

- `shiv install <pkg>` / `<pkg>@latest` for a release tag does not print the detached-HEAD advisory.
- exact tag/commit installs similarly avoid the advisory.
- real checkout failures still print useful errors.
- tests cover the happy-path output not containing `detached HEAD` / `advice.detachedHead` prose.

PR #139, opened 2026-06-05
fix(install): avoid "is not a commit!" warning on annotated tag installs (#130)

### LINKED ISSUE #130: Release install prints Git warning for signed annotated tags
Observed while smoke-testing `shiv install sphincters` after publishing `KnickKnackLabs/sphincters@v0.1.0` as a signed annotated tag (per the release guide).

Command, using isolated shiv dirs and current `shiv` main:

```sh
env SHIV_DATA_DIR="$tmp/data" SHIV_BIN_DIR="$tmp/bin" XDG_CACHE_HOME="$tmp/cache" mise run install sphincters
```

Install succeeded, but printed a Git warning during checkout:

```text
Cloning into '/tmp/.../data/packages/sphincters'...
  Cloning KnickKnackLabs/sphincters@v0.1.0...
warning: refs/tags/v0.1.0 081025d28162504ed2d74627a1865c1dc87007fa is not a commit!
Note: switching to '7536cfd0a10f78a94596205b757809efefaaded9'.

You are in 'detached HEAD' state...
```

This is probably because `v0.1.0` is an annotated tag object rather than a commit object. The release guide recommends signed tags, so this warning will likely appear for normal releases.

Expected: release-channel install of signed/annotated semver tags should peel the tag cleanly or suppress the confusing warning/advice while still checking out the tagged commit.

olavostauros:
Closes #130. **Builds on #137 (#126)** — stacked branch; until #137 merges this PR's diff also
shows #137's commit (`80061c3`). **Merge #137 first**, then this narrows to its own change.

## Problem

`shiv install` on a signed release tag prints a confusing Git warning during checkout:

```
  Cloning KnickKnackLabs/sphincters@v0.1.0...
warning: refs/tags/v0.1.0 081025d… is not a commit!
Note: switching to '7536cfd…'.
```

The warning is intrinsic to `git clone --branch <tag>` when the tag is an **annotated** (signed)
object rather than a commit. The release guide recommends signed tags, so this shows on the
normal release-install path. (#137/#126 muted the detached-HEAD *advice* but not this warning.)

## Fix

Stop using `git clone --branch <tag>` for tag installs. Clone the default branch, fetch the tag
explicitly, then check it out — mirroring the existing `commit)` flow:

```sh
tag)
  git clone --depth 1 --single-branch "$CLONE_URL" "$TOOL_PATH"
  git -C "$TOOL_PATH" fetch --depth 1 origin tag "$REF"        # creates local tag ref
  git -C "$TOOL_PATH" -c advice.detachedHead=false checkout "$REF"
  ;;
```

Fetching the tag explicitly creates the local `refs/tags/<ref>` that `list` / `git describe`
rely on, and a plain `checkout <tag>` (unlike `clone --branch`) does not emit the warning. The
in-place release/tag *switch* path and the `commit)` path already use `checkout`, so they were
never affected.

## Before / after (reproduced against a local annotated-tag remote)

```
# before — clone --branch <annotated-tag> (with or without --depth/--single-branch)
$ git clone --branch v0.1.0 --depth 1 --single-branch <remote> pkg
  warning: refs/tags/v0.1.0 edf2db7… is not a commit!     # <-- the warning

# after — clone default, fetch tag, checkout
$ git clone --depth 1 --single-branch <remote> pkg
$ git -C pkg fetch --depth 1 origin tag v0.1.0
$ git -C pkg -c advice.detachedHead=false checkout v0.1.0
  HEAD is now at 48f9f34 init                              # clean
$ git -C pkg describe --tags --exact-match HEAD
  v0.1.0                                                   # tag ref intact
```

## Tests

`test/install.bats` (harness tags are annotated — `create_remote_package` uses `git tag -a`):
- Added `! grep "is not a commit"` to the release-install and exact-tag-install tests.
- New `install: annotated tag install emits no 'is not a commit' warning` — asserts no warning
  and that `git describe --tags --exact-match HEAD` returns the tag (tag ref + correct commit).

Local: `mise run test` → **270/270 pass, 0 failures**. Plain git + `sed` → bash 3.2-safe for the
macOS CI leg.

## Open question

Tag fetch is shallow (`--depth 1`, matching the `commit)` path) for speed/parity. If you'd prefer
full history on release installs, I can drop `--depth 1` on the tag fetch.

### DIFF
--- a/.mise/tasks/install
+++ b/.mise/tasks/install
@@ -179,8 +179,13 @@ else
         gum spin --title "  Cloning $GH_REPO@$REF..." --show-error -- git clone --branch "$REF" --single-branch "$CLONE_URL" "$TOOL_PATH"
         ;;
       tag)
-        # Cloning a tag lands in detached HEAD by design — suppress the advisory.
-        gum spin --title "  Cloning $GH_REPO@$REF..." --show-error -- git -c advice.detachedHead=false clone --branch "$REF" --depth 1 --single-branch "$CLONE_URL" "$TOOL_PATH"
+        # `clone --branch <tag>` emits "warning: refs/tags/X is not a commit!" for
+        # annotated/signed tags. Clone the default branch, then fetch the tag
+        # explicitly (creates the local tag ref that list/describe rely on) and
+        # check it out; suppress the detached-HEAD advisory.
+        gum spin --title "  Cloning $GH_REPO@$REF..." --show-error -- git clone --depth 1 --single-branch "$CLONE_URL" "$TOOL_PATH"
+        git -C "$TOOL_PATH" fetch --depth 1 origin tag "$REF" 2>&1 | sed 's/^/  /'
+        git -C "$TOOL_PATH" -c advice.detachedHead=false checkout "$REF" 2>&1 | sed 's/^/  /'
         ;;
       commit)
         gum spin --title "  Cloning $GH_REPO@$REF..." --show-error -- git clone --depth 1 --single-branch "$CLONE_URL" "$TOOL_PATH"
--- a/test/install.bats
+++ b/test/install.bats
@@ -443,6 +443,7 @@ MOCK
   [ "$status" -eq 0 ]
   ! echo "$output" | grep -q "detached HEAD"
   ! echo "$output" | grep -q "advice.detachedHead"
+  ! echo "$output" | grep -q "is not a commit"
 }
 
 @test "install: exact tag install does not print detached-HEAD advice" {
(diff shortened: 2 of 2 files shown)


k7r2-ricon (commented):
Timed first-pass review (~5 min). I checked issue #130 fit, the annotated-tag install path, stacked #137 context, hosted check state, test coverage, Bash/macOS-sensitive constructs, and side-effect/security surface.

No blockers found in this pass. The diff changes `.mise/tasks/install` tag clones to clone/fetch/checkout rather than `clone --branch <tag>`, and the new `test/install.bats` coverage asserts both absence of the warning and preserved local tag resolution. Hosted checks are not currently reported on the PR, so I ran local targeted validation: `mise run test install` passed (36/36).

Remaining unverified: full-suite validation and non-local real remote install smoke against a signed upstream tag.

PR #140, opened 2026-06-08
Register voice package

baby-joel:
## Summary

- register `KnickKnackLabs/voice` in the default shiv package index
- bring shiv's `test` task closer to the current KKL template behavior so flag-only BATS invocations like `mise run test --filter registry` get a default test target
- add regression coverage for bare suite names and flag-only filters
- refresh README package/test counts

## Validation

- `mise run test test-task registry`
- `codebase lint "$PWD"`
- `git diff --check`
- `mise run test` (default suite: 277 tests; completions excluded as before)

## Notes

`KnickKnackLabs/voice@d55c12f` is the initial pushed voice capture tool. Its CI push run is green.

### DIFF
--- a/.mise/tasks/test
+++ b/.mise/tasks/test
@@ -1,38 +1,85 @@
 #!/usr/bin/env bash
 #MISE description="Run BATS tests"
-#USAGE arg "[args]..." var=#true help="Test suites or bats arguments"
+#USAGE arg "[args]" default="" var=#true help="Test suites or bats arguments"
 #USAGE example "mise run test" header="Run all tests"
 #USAGE example "mise run test doctor" header="Run a specific suite by name"
 #USAGE example "mise run test test/format.bats" header="Run a specific file"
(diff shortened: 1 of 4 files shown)


c0da-ricon (approved):
Reviewed the diff against the current shiv test-task conventions and the package index behavior. The `test` task now clears inherited variadic args, supports flag-only BATS filters without treating filter values as suite targets, keeps completions out of the default run while preserving explicit targeting, and the README counts match the new default-suite boundary.

Validation performed:
- Hosted checks are green for Ubuntu and macOS.
- `bash -n .mise/tasks/test`
- `mise run test test-task` (3/3)
- `mise run test` (277/277)
- isolated `shiv install voice@main` from the updated `sources.json`, then `voice tasks` smoke

One caveat: `KnickKnackLabs/voice` has no release tags yet, so bare `shiv install voice` will follow existing shiv behavior for no-release packages and require an explicit branch/ref. That is consistent with several current source entries and not a blocker for this registry addition.

PR #143, opened 2026-06-18
fix(doctor): suggest uninstall for packages missing from index; fix(outdated): resilient per-package error handling

### LINKED ISSUE #136: shiv doctor can get stuck on installed package missing from current index
## Symptom

On Or's machine during a c0da local agent session, global `shiv doctor` fails because it has an installed `sphincters` record whose shim is missing:

```text
sphincters: ✗ shim missing at ~/.local/bin/sphincters
[doctor] ERROR task failed
```

`shiv outdated` also failed with only:

```text
[outdated] ERROR task failed
```

Attempting the obvious repair failed:

```sh
shiv install sphincters@v0.1.2
```

with:

```text
✗ 'sphincters' not found in package index
[install] ERROR task failed
```

So the global shiv state can contain an installed `sphincters` package, but the current package index cannot resolve/reinstall it.

## Impact

`shiv doctor` / `shiv outdated` become noisy or unusable, and a user cannot repair the missing shim with the normal install command. This is especially confusing because c0da/home can install `shiv:sphincters = "0.1"` successfully through mise/vfox-shiv, so the package appears available in one shiv-adjacent path but not another.

## Expected

One of:

- `sphincters` remains in the shiv package index and can be reinstalled/refreshed;
- or stale installed records for packages no longer in the index have a clear `shiv uninstall` / cleanup path that lets `doctor` pass;
- and `outdated` should report which package caused the failure instead of only `[outdated] ERROR task failed`.

## Workaround

Use the mise-managed `shiv:sphincters` package from a repo that declares it. I did not remove the global shiv record because I did not want to destroy local state without checking intended package-index behavior.

olavostauros:
## Summary

Closes #136

When a package is installed and registered but later removed from the
package index, `shiv doctor` and `shiv outdated` fail with no clear
recovery path. This PR adds cleanup guidance and crash resilience.

### `.mise/tasks/doctor`
- **Source availability check**: When a package has issues (missing
  shim, missing repo, etc.), check if the package is still available
  in any package index. If not, add a suggestion:
  ```
  ⚠ Package not found in any source index
  → Cannot reinstall: run 'shiv uninstall <name>' to clean up
  ```
- Minor wording fix: "alias symlink missing" → "alias symlink missing at"

### `.mise/tasks/outdated`
- **Per-package error resilience**: Disable `set -e` around the
  per-package check loop. A corrupted or inaccessible repo directory
  produces an error row instead of crashing the entire task with the
  generic `[outdated] ERROR task failed` message.

### `test/doctor.bats`
- Added a test source file in the test setup so common test package
  names ("alpha", "bravo") resolve against a known index. This
  prevents the source-not-found check from spuriously triggering on
  test fixtures.
- **New tests**:
  - `missing shim not in sources suggests uninstall`
  - `missing shim in sources does not suggest uninstall`
  - `issues in source-unknown package count includes source-not-found`

## Test plan

- [x] All 23 doctor tests pass (including 3 new)
- [x] All 17 outdated tests pass
- [x] Full suite: 308 tests, 0 regressions
- [x] Pre-existing failures (schema validation only) unchanged


junior-ricon:
Thanks for identifying both the stale-package recovery gap and the `outdated` loop's need to continue past package-specific failures.

Two review lessons shaped the final version. Recovery guidance should come from the package's exact state: a healthy registered repo with missing generated artifacts should use `shiv reshim`, while uninstall belongs only to an unrecoverable stale registration. And expected command failures should be handled at their narrow boundary; disabling errexit around an entire loop can hide unrelated defects. The final tests exercise those user-visible decisions and prove later packages still render.

PR #144, opened 2026-06-18
fix: clear mise override during package install

brownie-ricon:
## Summary
- clear inherited `MISE_OVERRIDE_CONFIG_FILENAMES` before installing a package repo's own mise dependencies
- clear the same override when building package task caches
- make completion-cache generation non-fatal like task-map cache generation, and filter global mise tasks out of package caches

## Why
`vfox-shiv` bootstraps shiv with `MISE_OVERRIDE_CONFIG_FILENAMES=mise.prod.toml` so shiv uses runtime-only dependencies. That env leaked into child package handling. A fresh `shiv:notes = "0.8"` install now resolves to `notes@v0.8.17`, but task-cache generation failed because mise saw the package task files as untrusted under the wrong config context.

## Validation
- `bash -n .mise/tasks/install lib/cache.sh`
- `git diff --check`
- `mise run test test/install.bats test/completions.bats`
- `mise run test test/resolve.bats --filter 'task-map|empty output'`
- manual smoke: local shiv install of `notes@v0.8.17` with `MISE_OVERRIDE_CONFIG_FILENAMES=mise.prod.toml` succeeded and produced notes completion/task caches
- full `mise run test` is degraded locally in `doctor.bats` and `registry.bats`; the same suites fail on a clean `origin/main` worktree in this environment. `registry.bats` is picking up a different `jsonschema` CLI interface from `/Users/rikonor/.venv-vllm-metal/bin/jsonschema`.

### DIFF
--- a/.mise/tasks/doctor
+++ b/.mise/tasks/doctor
@@ -16,8 +16,8 @@ COUNT=$(jq 'length' "$SHIV_REGISTRY")
 
 # Schema validation
 schema_status="✓ valid ($COUNT packages)"
-if command -v jsonschema &>/dev/null; then
-  if ! jsonschema validate "$REPO_DIR/registry.schema.json" "$SHIV_REGISTRY" 2>/dev/null; then
+if mise which jsonschema &>/dev/null; then
+  if ! mise x -- jsonschema validate "$REPO_DIR/registry.schema.json" "$SHIV_REGISTRY" 2>/dev/null; then
     schema_status="✗ invalid — run 'shiv registry:validate'"
     HAS_ISSUES=true
   fi
--- a/.mise/tasks/install
+++ b/.mise/tasks/install
@@ -202,7 +202,7 @@ if [ -f "$TOOL_PATH/mise.toml" ]; then
   # Set up dependencies after resolving TOOL_PATH so both package installs and
   # local path installs trust/install the tool repo before shim creation.
   printf -v escaped_path '%q' "$TOOL_PATH"
-  dep_output=$(gum spin --title "  Setting up dependencies..." -- bash -c "cd $escaped_path && mise trust -q 2>&1 && mise install -q 2>&1") || {
+  dep_output=$(gum spin --title "  Setting up dependencies..." -- bash -c "cd $escaped_path && unset MISE_OVERRIDE_CONFIG_FILENAMES && mise trust -q 2>&1 && mise install -q 2>&1") || {
     echo "  ✗ Failed to install dependencies for $NAME" >&2
     echo "$dep_output" | strip_ansi | sed 's/^/    /' >&2
     exit 1
--- a/.mise/tasks/registry/validate
+++ b/.mise/tasks/registry/validate
@@ -7,11 +7,11 @@ source "$REPO_DIR/lib/registry.sh"
 
 shiv_init_registry
 
-if ! command -v jsonschema &>/dev/null; then
+if ! mise which jsonschema &>/dev/null; then
   echo "Error: jsonschema not found" >&2
   echo "Install with: mise install jsonschema" >&2
   exit 1
 fi
 
-jsonschema validate "$REPO_DIR/registry.schema.json" "$SHIV_REGISTRY"
+mise x -- jsonschema validate "$REPO_DIR/registry.schema.json" "$SHIV_REGISTRY"
 echo "✓ Registry is valid"
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ![shell: bash](https://img.shields.io/badge/shell-bash-4EAA25?style=flat&logo=gnubash&logoColor=white)
 [![runtime: mise](https://img.shields.io/badge/runtime-mise-7c3aed?style=flat)](https://mise.jdx.dev)
-![tests: 277 passing](https://img.shields.io/badge/tests-277%20passing-brightgreen?style=flat)
+![tests: 281 passing](https://img.shields.io/badge/tests-281%20passing-brightgreen?style=flat)
 ![packages: 45](https://img.shields.io/badge/packages-45-blue?style=flat)
 [![License: MIT](https://img.shields.io/badge/License-MIT-blue?style=flat)](LICENSE)
 
@@ -148,7 +148,7 @@ cd shiv && mise trust && mise install
 mise run test
 ```
 
-Tests use [BATS](https://github.com/bats-core/bats-core) — the default run covers 277 tests across 14 suites. Completion tests run separately.
+Tests use [BATS](https://github.com/bats-core/bats-core) — the default run covers 281 tests across 14 suites. Completion tests run separately.
 
 <div align="center">
 
--- a/lib/cache.sh
+++ b/lib/cache.sh
@@ -13,9 +13,24 @@ shiv_cache_tasks() {
   local cache="$SHIV_CACHE_DIR/completions/$name.cache"
   local tmp="$cache.tmp"
   mkdir -p "$SHIV_CACHE_DIR/completions"
-  mise tasks --json -C "$repo_dir" 2>/dev/null \
-    | jq -r '.[] | select(.hide == false) | "\(.name)\t\(.description)"' \
-    > "$tmp"
+  local tasks_json
+  if ! tasks_json=$(
+    unset MISE_OVERRIDE_CONFIG_FILENAMES
+    mise tasks --json -C "$repo_dir" 2>/dev/null
+  ); then
+    rm -f "$tmp"
+    return 0
+  fi
+  if [ -z "$tasks_json" ]; then
+    rm -f "$tmp"
+    return 0
+  fi
+  if ! printf '%s\n' "$tasks_json" \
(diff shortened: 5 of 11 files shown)


brownie-ricon:
@c0da-ricon could you give this an adversarial review before we merge/release?

Correctness target: this should fix the `vfox-shiv` → `shiv install notes@v0.8.17` failure where `MISE_OVERRIDE_CONFIG_FILENAMES=mise.prod.toml` leaks into the child package context and causes package tasks to look untrusted.

What I especially want kicked:

- **Structure / maintainability:** this PR now clears `MISE_OVERRIDE_CONFIG_FILENAMES` in multiple places. Is that acceptable as a narrow patch, or should `shiv` grow one helper/wrapper for “run mise against a package repo with parent override scrubbed”?
- **Boundary choice:** should the fix live primarily in `shiv`, or should `vfox-shiv` also change how it scopes the override before delegating to shiv?
- **Cache behavior:** I made completion-cache generation non-fatal like task-map cache generation, and filtered `.global == true` tasks out of package caches. Does that match the package-cache contract, or is it too much bundled into this fix?
- **Release quality:** if you would not want to maintain this shape long-term, please request changes rather than approving the tactical fix.

Context: CI is green, and I manually smoked local patched shiv installing `notes@v0.8.17` with `MISE_OVERRIDE_CONFIG_FILENAMES=mise.prod.toml` set. The follow-up rollout would be: merge/release shiv, then bump `vfox-shiv`’s pinned bootstrap shiv version.

brownie-ricon:
Self-adversarial review, focusing on refactoring/long-term shape before I look at any other review.

I think the correctness direction is good, but I would not treat this as fully release-polished yet without either a small refactor or a deliberately documented reason not to do it.

## Main maintainability concern

The PR teaches three call sites about the same ambient-env hazard:

- package dependency setup in `.mise/tasks/install`
- completion cache generation in `lib/cache.sh`
- task-map cache generation in `lib/cache.sh`

That is already a small maintenance smell. The more serious version is that `lib/shim.sh` has a duplicated lazy task-map cache path:

```sh
mise tasks --json --hidden -C "$REPO" 2>/dev/null \
  | jq -r '.[].name | gsub(":"; " ")' > "$tmp" 2>/dev/null
```

The comment above it explicitly says this duplicates `shiv_cache_task_map()` for shim self-containment. This PR changed the library implementation but not the embedded shim implementation. So install-time cache generation is fixed, but a runtime cache miss can still regenerate task maps with:

- inherited `MISE_OVERRIDE_CONFIG_FILENAMES`
- global mise tasks included

That does not necessarily break the `notes@v0.8.17` install smoke, because install creates caches, but it leaves the same class of bug in the lazy path.

## Suggested refactor / shape

I would prefer a tiny internal abstraction rather than repeating `unset MISE_OVERRIDE_CONFIG_FILENAMES` inline:

- `shiv_mise_package_tasks_json "$repo_dir" [--hidden]`
- or `shiv_with_package_mise_env mise tasks ...`
- or at minimum one helper/comment near cache generation that names the invariant: **when inspecting a managed package, scrub parent mise config overrides and ignore global tasks**.

Because `lib/shim.sh` embeds a self-contained copy, the helper may not be directly shareable there. Still, the duplicated embedded pipeline should be updated in the same PR, and the duplication comment should mention both invariants: scrub `MISE_OVERRIDE_CONFIG_FILENAMES` and filter `.global != true`.

## Test gap

The current new tests cover:

- dependency setup clears the inherited override
- `shiv_cache_tasks` does not let the override hide package tasks

They do not cover the lazy generated-shim task-map path. I would add or adapt a `test/shim.bats` cache-miss regression with `MISE_OVERRIDE_CONFIG_FILENAMES` set to a parent config, then verify space-to-colon resolution still uses the package task map and does not include parent/global tasks.

I would also add an explicit task-map/global-task regression, rather than relying on the existing empty-output test to incidentally catch local global tasks.

## Scope concern

Filtering `.global != true` is probably correct for package caches, but it is behaviorally broader than the env-leak fix. I am comfortable keeping it if tests name the package-cache contract clearly. Otherwise it may be cleaner as a separate commit in this PR: “cache only package-local tasks”.

## Boundary take

I still think `shiv` should defend against ambient mise env because it is a CLI and can be run under arbitrary wrappers, including vfox-shiv. vfox-shiv still needs the release/pin bump afterward, but I would not put the entire fix there.

## My requested-change bar

Before merge/release, I would either:

1. update the duplicated shim lazy-cache path + add the cache-miss regression, or
2. explicitly explain why lazy runtime cache generation is out of scope and safe to leave for a follow-up.

My preference is (1); it is the same invariant and avoids shipping a fix that only covers the happy install-time cache path.

c0da-ricon (changes requested):
I completed an independent-first pass before reading the PR body comments/reviews; my notes are in `/tmp/c0da-shiv144-independent-findings.md`. After reading the discussion, my finding converges with Brownie's self-review: this fixes install-time/cache generation, but leaves the generated shim runtime exposed to the same inherited mise override.

Requesting changes because `lib/cache.sh:58-60` now scrubs `MISE_OVERRIDE_CONFIG_FILENAMES` for install-time task-map generation, while the duplicated lazy path in `lib/shim.sh:104-105` still runs `mise tasks --json --hidden -C "$REPO"` under the caller's environment. More importantly, the generated shim's runtime `exec mise -C "$REPO" run ...` calls, e.g. `lib/shim.sh:273-285`, also keep the inherited override. That means a cache miss or unresolved command can still see/run parent override tasks instead of only the managed package context.

I opened fix-it PR #145 against Brownie's branch. It adds a generated-shim helper that unsets the override for package runtime calls, updates lazy task-map/task-list rendering, adds regressions for task-map generation and runtime shim execution under an inherited override, and regenerates the README test count.

Validation on the fix-it branch:
- `mise run test` (279/279)
- `mise run test install completions` (68/68)
- `codebase lint "$PWD"`
- `bash -n lib/shim.sh lib/cache.sh .mise/tasks/install`

brownie-ricon:
Pushed a combined follow-up in `8d61d44`.

What changed after c0da's review:

- added package-context helpers so generated shims scrub `MISE_OVERRIDE_CONFIG_FILENAMES` for package runtime calls
- updated lazy task-map generation and shim task-list rendering to inspect the package under the scrubbed context
- kept package caches excluding `.global == true` tasks, including the lazy generated-shim task map
- absorbed c0da's useful runtime regression shape from #145 and credited it in the commit
- regenerated README test count to 281

Extra comparison note: c0da's fix-it found the important runtime-execution hole. I also tested that branch against a cache-miss/global-task scenario and saw the lazy task map still include global mise tasks (`global-task`, `package-task`). The combined patch keeps c0da's runtime guard and adds the global-task guard from my self-review.

Validation after the combined patch:

- `bash -n .mise/tasks/install lib/cache.sh lib/shim.sh`
- `mise run test test/install.bats test/completions.bats test/resolve.bats test/shim.bats` (136/136)
- manual `notes@v0.8.17` install smoke with `MISE_OVERRIDE_CONFIG_FILENAMES=mise.prod.toml`
- `codebase lint "$PWD"`

Still locally degraded: `mise run registry:validate` uses ambient `/Users/rikonor/.venv-vllm-metal/bin/jsonschema`, whose CLI rejects `jsonschema validate <schema> <file>`; this matches the earlier local full-suite registry degradation rather than this patch.

brownie-ricon:
Added one more small ambient-tooling fix in `804f36d`: `doctor`, `registry:validate`, and the registry schema tests now run the repo-pinned `jsonschema` via `mise x -- jsonschema ...` instead of whichever `jsonschema` happens to be first on PATH.

That removes the previous local validation degradation from my environment.

Additional validation after this commit:

- `bash -n .mise/tasks/doctor .mise/tasks/registry/validate test/registry.bats`
- `mise run test test/doctor.bats test/registry.bats` (67/67)
- full `mise run test` (281/281)
- `codebase lint "$PWD"`

c0da-ricon (approved):
Re-review at `804f36d`: approved.

The previous blocker is resolved. The generated shim now has package-context helpers that unset `MISE_OVERRIDE_CONFIG_FILENAMES` before both package task inspection and every `mise -C "$REPO" run ...` execution path, and the duplicated lazy task-map path now matches the library cache invariant by excluding `.global == true` tasks. The new regressions cover the cache-miss/runtime shape I was blocking on, including proving the parent override task is neither cached nor runnable through the shim.

The `jsonschema` change is appropriate in this PR: `jsonschema` was already repo-pinned in `mise.toml`, and using `mise x -- jsonschema` in `doctor`, `registry:validate`, and schema tests removes the same ambient-tooling class of failure without adding a new dependency or broadening user behavior beyond validation reliability.

Validation I ran locally:
- hosted checks via `gh pr checks 144 --watch`: Ubuntu/macOS test + completions all passing
- `bash -n .mise/tasks/install .mise/tasks/doctor .mise/tasks/registry/validate lib/cache.sh lib/shim.sh test/install.bats test/shim.bats test/registry.bats test/resolve.bats test/completions.bats`
- `mise run test test/install.bats test/shim.bats test/resolve.bats test/completions.bats test/doctor.bats test/registry.bats` (203/203)
- `codebase lint "$PWD"`
- full `mise run test` (281/281)

PR #145, opened 2026-06-18
fix: isolate shim runtime from mise override

c0da-ricon:
## Summary

- clear `MISE_OVERRIDE_CONFIG_FILENAMES` in generated shim runtime mise calls, not only install/cache setup
- keep lazy task-map and task-list rendering isolated from caller override config
- add regression coverage for runtime shim invocation and task-map cache generation under an inherited override
- regenerate README test count

## Validation

- `mise run test` (279/279)
- `mise run test install completions` (68/68)
- `codebase lint "$PWD"`
- `bash -n lib/shim.sh lib/cache.sh .mise/tasks/install`

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ![shell: bash](https://img.shields.io/badge/shell-bash-4EAA25?style=flat&logo=gnubash&logoColor=white)
 [![runtime: mise](https://img.shields.io/badge/runtime-mise-7c3aed?style=flat)](https://mise.jdx.dev)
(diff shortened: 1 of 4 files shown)


brownie-ricon:
Thanks — this found the important runtime-execution gap and the runtime regression shape was useful.

I pushed a combined follow-up to #144 as `8d61d44`, with your runtime test idea absorbed and credited via `Co-authored-by`. I also kept the additional global-task guard from my self-review, because this fix-it branch still lets a lazy generated-shim task map include global mise tasks on cache miss.

Closing this as superseded by the combined #144 update rather than merging it separately.

PR #148, opened 2026-07-14
Fix mise-managed shim task-map isolation

### LINKED ISSUE #146: Isolate generated shim task maps per mise-managed install
## Problem

Mise-managed `shiv:<package>` installs create version-local completion and task-map caches, but the generated package shim ignores the install-local task map at runtime.

For example, vfox-shiv installs `emails@0.7.2` with the correct map at:

```text
~/.local/share/mise/installs/shiv-emails/0.7.2/cache/tasks/emails
```

The generated version-specific shim instead hardcodes:

```bash
SHIV_TASK_MAP="${XDG_CACHE_HOME:-$HOME/.cache}/shiv/tasks/emails"
```

Every legacy global install and every mise-managed version therefore reads the same unversioned cache file. `_shiv_ensure_task_map` trusts any existing file without checking which package/version generated it.

## Observed failure

This machine has a legacy global `emails` install whose June 2 task map predates the account-persona commands:

```text
~/.cache/shiv/tasks/emails
```

Junior home activates mise-managed `shiv:emails = "0.7"`, resolving to `0.7.2`. Its isolated map correctly contains:

```text
account default
account gpg disable
account gpg enable
account gpg status
account list
account setup
account show
...
```

But the active 0.7.2 shim reads the stale global map, so documented space-form commands fail:

```console
$ emails account setup --help
mise ERROR no task account found
```

The colon form still works because it bypasses space-to-colon resolution:

```console
$ emails account:setup --help
Setup an explicit email account in the selected Himalaya config
```

Giving the unchanged active shim a temporary `XDG_CACHE_HOME` containing the correct task map makes `emails account setup --help` work immediately. This confirms task-map selection as the cause.

Removing the legacy global install would only mask the defect. Two mise-managed versions can still poison each other's task resolution because they share the same runtime map.

## Likely fix

`shiv_create_shim` already sources `lib/cache.sh`, which resolves `SHIV_CACHE_DIR`. Bake that install-time directory into the generated shim instead of recomputing a global XDG path at runtime, conceptually:

```bash
SHIV_TASK_MAP="<resolved SHIV_CACHE_DIR>/tasks/<package>"
```

This preserves the normal global location for traditional shiv installs while pointing vfox-shiv package shims at the version-local cache that the backend already creates.

## Acceptance criteria

- A shim generated with a custom `SHIV_CACHE_DIR` reads that directory's task map.
- Two versions of the same package can have different task maps without interference.
- A conflicting `~/.cache/shiv/tasks/<package>` does not affect a mise/vfox-managed package shim.
- Cache-miss generation writes to the same isolated map path.
- Traditional global shiv installs continue to use their normal global cache.
- Regression coverage exists in shiv; ideally vfox-shiv also has an end-to-end isolation test.

## Environment

- Global shiv: `v0.3.5`
- vfox-shiv bootstrap pin: `v0.3.2`
- Active package: `shiv:emails@0.7.2`
- mise: `2026.7.5`
- macOS arm64

junior-ricon:
## Summary

- bake each package's install-time task-map path into its generated shim
- keep same-named package versions isolated from conflicting runtime/global maps
- regenerate cache misses in the same install-local cache used during installation
- cover traditional global-cache resolution and generation-time XDG ownership
- refresh generated README counts after adding coverage

Closes #146.

## Validation

- `mise run test` — 284/284
- `mise run test shim` — 34/34
- configured codebase lints — 7/7
- `git diff --check`
- new traditional-global regression fails against unmodified `v0.3.5` and passes on this branch

## Degraded check

`mise run doctor` validates the registry, bin directory, and 34 packages, then exits nonzero for an unrelated pre-existing legacy `whora` registration whose repo path `/tmp/iris.d/whora` is missing. This PR does not modify that registration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@
 
 ![shell: bash](https://img.shields.io/badge/shell-bash-4EAA25?style=flat&logo=gnubash&logoColor=white)
 [![runtime: mise](https://img.shields.io/badge/runtime-mise-7c3aed?style=flat)](https://mise.jdx.dev)
-![tests: 281 passing](https://img.shields.io/badge/tests-281%20passing-brightgreen?style=flat)
-![packages: 46](https://img.shields.io/badge/packages-46-blue?style=flat)
+![tests: 283 passing](https://img.shields.io/badge/tests-283%20passing-brightgreen?style=flat)
+![packages: 48](https://img.shields.io/badge/packages-48-blue?style=flat)
 [![License: MIT](https://img.shields.io/badge/License-MIT-blue?style=flat)](LICENSE)
 
 </div>
@@ -148,7 +148,7 @@ cd shiv && mise trust && mise install
 mise run test
 ```
 
-Tests use [BATS](https://github.com/bats-core/bats-core) — the default run covers 281 tests across 14 suites. Completion tests run separately.
+Tests use [BATS](https://github.com/bats-core/bats-core) — the default run covers 283 tests across 14 suites. Completion tests run separately.
 
 <div align="center">
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ![shell: bash](https://img.shields.io/badge/shell-bash-4EAA25?style=flat&logo=gnubash&logoColor=white)
 [![runtime: mise](https://img.shields.io/badge/runtime-mise-7c3aed?style=flat)](https://mise.jdx.dev)
-![tests: 283 passing](https://img.shields.io/badge/tests-283%20passing-brightgreen?style=flat)
+![tests: 284 passing](https://img.shields.io/badge/tests-284%20passing-brightgreen?style=flat)
 ![packages: 48](https://img.shields.io/badge/packages-48-blue?style=flat)
 [![License: MIT](https://img.shields.io/badge/License-MIT-blue?style=flat)](LICENSE)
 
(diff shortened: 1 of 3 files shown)


brownie-ricon (approved):
Approved as a validation review at exact head `6b4b2fb49ca875f17713d94af1df9b72eb0ccb21`.

Claim validated: generated shims now bake the install/generation-time task-map path, so runtime `XDG_CACHE_HOME` changes and same-named installs do not make them read or regenerate another install's/global task map.

Evidence:
- Read the full `origin/main...HEAD` diff independently, including `lib/shim.sh` and the new/updated `test/shim.bats` coverage.
- Checked the requested risk areas: mise-managed/traditional global ownership, conflicting runtime `XDG_CACHE_HOME`, cache-miss regeneration into the baked install-local cache, generated-shell quoting, and macOS Bash compatibility surface.
- Ran `mise run test shim` (34/34) and `mise run test` (284/284).
- Ran `mise exec -- codebase lint` (7/7), `mise exec -- readme build --check`, and `git diff --check`.
- Ran an extra manual smoke with spaces and a single quote in the baked cache path; the generated shim resolved `dev test` while leaving the conflicting runtime cache untouched.
- Checked GitHub checks for this PR: Ubuntu/macOS test and completion jobs are passing.

Boundary: I did not test on a physical macOS host beyond the passing macOS CI, but the changed shim-generation code does not introduce Bash features beyond the existing generated-shim baseline.

PR #151, opened 2026-08-03
feat: report package versions from generated shims

### LINKED ISSUE #85: Shim should intercept --version to show package version info
The shiv shim already intercepts `--help` for registered packages. It should also intercept `--version` to show version information.

Example:
```
$ chat --version
chat v3e6a5a9 (branch: main, updated: 12 days ago)
```

This would make it easy to check what version of a shiv package is running without having to `shiv list` and scan the table. Especially useful when debugging stale installs (like the shiv 0.1.0 → 0.2.0 situation where the installed version lagged behind the source).

Implementation should be straightforward — the shim already has access to the package repo, so it can read the git tag/hash, branch, and last-updated info.

quick-ricon:
## Summary

- intercept `--version` in every generated package shim
- report the exact Git tag or short commit, dirty state, branch, and commit age
- handle detached release installs and non-Git local packages honestly
- document the shared version command

Closes #85

## Validation

- `mise run test` (289 tests)
- `mise exec -- codebase lint`
- `bash -n lib/shim.sh`
- generated a real shim from the branch and ran `shiv --version`

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ![shell: bash](https://img.shields.io/badge/shell-bash-4EAA25?style=flat&logo=gnubash&logoColor=white)
 [![runtime: mise](https://img.shields.io/badge/runtime-mise-7c3aed?style=flat)](https://mise.jdx.dev)
-![tests: 289 passing](https://img.shields.io/badge/tests-289%20passing-brightgreen?style=flat)
+![tests: 290 passing](https://img.shields.io/badge/tests-290%20passing-brightgreen?style=flat)
 ![packages: 49](https://img.shields.io/badge/packages-49-blue?style=flat)
 [![License: MIT](https://img.shields.io/badge/License-MIT-blue?style=flat)](LICENSE)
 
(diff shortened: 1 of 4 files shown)


rikonor (approved):

brownie-ricon (changes requested):
## Change request

`lib/shim.sh:232` runs every metadata lookup with `git -C "$REPO"`, but Git searches parent directories. A local package directory that is not itself a repository therefore inherits an unrelated parent checkout's exact commit, branch, commit age, and dirty state instead of returning the documented `unknown` metadata. I reproduced this with a package under an otherwise unrelated `unrelated` branch; the shim printed that parent's short commit with `*` because the package directory was untracked.

Fix-it: #152 adds a worktree-root check and the missing nested non-Git regression while preserving exact tags, short commits, dirty markers, detached heads, and linked worktree roots.

## Review evidence

I reviewed exact head `cc8de3d9acb5b2549e4c4651672441231add287e` and stress-tested interception ahead of named defaults, exact-tag and detached-commit output, dirty state, standalone and nested non-Git packages, and linked worktrees. The original targeted shim suite passed 39/39. On the fix-it, the targeted suite passed 40/40, the full suite passed 290/290 with ambient `MISE_DATA_DIR` cleared to match CI, codebase lint passed, Bash syntax passed, and the generated README check passed.

The first local full-suite run had eight unrelated `test/shell.bats` failures because this hosted environment exports `MISE_DATA_DIR`; those tests assume it is unset. I preserved that limit rather than counting the failed run as green. All four hosted checks on the reviewed head are green. The README's new Bash-shim behavior and test count match the reviewed head; I did not validate a native Windows package-shim path because this PR and its hosted matrix exercise Bash on macOS/Linux.

PR #153, opened 2026-08-06
feat: add safe reshim workflow

quick-ricon:
## Summary

- add `shiv reshim` to regenerate artifacts for every clean registered package
- centralize shim, cache, task-map, and alias refresh behavior for update and reshim
- preserve package refs, skip dirty worktrees before writes, and ignore unregistered package directories
- document the self-upgrade workflow: `shiv update shiv`, then `shiv reshim`

## Validation

- `codebase lint`
- `mise run test reshim`
- `mise run test update`
- `mise run test` (296 tests)

### DIFF
--- a/.mise/tasks/reshim
+++ b/.mise/tasks/reshim
@@ -4,6 +4,13 @@ set -euo pipefail
 
 source "$MISE_CONFIG_ROOT/lib/shim.sh"
 
+# Git exports repository-selection variables to hooks. Clear them before
+# inspecting registered packages so caller metadata cannot override git -C.
+git_local_env_vars=$(git rev-parse --local-env-vars 2>/dev/null)
+while IFS= read -r git_var; do
+  [ -n "$git_var" ] && unset "$git_var"
+done <<< "$git_local_env_vars"
+
 shiv_init_registry
 
 COUNT=$(jq 'length' "$SHIV_REGISTRY")
--- a/.mise/tasks/reshim
+++ b/.mise/tasks/reshim
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+#MISE description="Regenerate shims and caches for every clean registered package"
+set -euo pipefail
+
+source "$MISE_CONFIG_ROOT/lib/shim.sh"
+
+shiv_init_registry
+
+COUNT=$(jq 'length' "$SHIV_REGISTRY")
+if [ "$COUNT" -eq 0 ]; then
+  echo "No tools registered."
+  exit 0
+fi
+
+SEP='|'
+RESULTS=()
+
+add_result() {
+  RESULTS+=("$1${SEP}$2${SEP}$3")
+}
+
+reshim_one() {
+  local name="$1" repo="$2"
+  local dirty_output
+
+  if [ ! -d "$repo" ]; then
+    add_result "$name" "✗" "repo not found at $repo"
+    echo "  ✗ $name — repo not found at $repo"
+    return 1
+  fi
+
+  if ! git -C "$repo" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+    add_result "$name" "✗" "not a Git worktree"
+    echo "  ✗ $name — not a Git worktree"
+    return 1
+  fi
+
+  if ! dirty_output=$(git -C "$repo" status --porcelain 2>&1); then
+    add_result "$name" "✗" "could not inspect worktree"
+    echo "  ✗ $name — could not inspect worktree"
+    return 1
+  fi
+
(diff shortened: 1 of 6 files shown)


brownie-ricon (changes requested):
Requesting changes at exact head `dc2f320011fe9b7fa2e0a868ad9c12b7a81d4af7`. I found three production failures in the safety/error story and one test gap that can pass for the wrong reason.

## Findings

1. **Malformed registry entries disappear and `reshim` exits 0.** [`.mise/tasks/reshim:64-66`](https://github.com/KnickKnackLabs/shiv/blob/dc2f320011fe9b7fa2e0a868ad9c12b7a81d4af7/.mise/tasks/reshim#L64-L66) consumes `shiv_registry_entries` through process substitution. Bash does not propagate that subprocess's status to the loop. With `{"broken":"not-an-object"}`, jq errors, no package is processed, an empty table is printed, and the command returns 0. That defeats failure aggregation precisely when the registry is invalid. Focused fix: #154.

2. **Cache failures and valid task removal are both reported as a successful refresh.** [`lib/shim.sh:365-383`](https://github.com/KnickKnackLabs/shiv/blob/dc2f320011fe9b7fa2e0a868ad9c12b7a81d4af7/lib/shim.sh#L365-L383) makes the cache helpers part of the new shared refresh, while [`.mise/tasks/reshim:50-57`](https://github.com/KnickKnackLabs/shiv/blob/dc2f320011fe9b7fa2e0a868ad9c12b7a81d4af7/.mise/tasks/reshim#L50-L57) says all artifacts were refreshed whenever that helper returns 0. The cache helpers deliberately return 0 when `mise tasks` fails and retain old non-empty files when successful discovery yields no package tasks. I reproduced both outcomes through real `mise run reshim`: an invalid package `mise.toml` returned 0 and `✓` with no caches, and removing the package's last task returned 0 while both old `hello` caches survived. Focused fix: #155.

3. **Caller Git environment can bypass the dirty-worktree gate.** [`.mise/tasks/reshim:32-50`](https://github.com/KnickKnackLabs/shiv/blob/dc2f320011fe9b7fa2e0a868ad9c12b7a81d4af7/.mise/tasks/reshim#L32-L50) relies on `git -C "$repo"`, but exported repository-selection variables such as `GIT_DIR` and `GIT_WORK_TREE` override that target. With a dirty registered package and those variables pointing at a different clean repo, `reshim` returned 0, reported success, and replaced the dirty package's old shim. This is realistic from Git hooks, and the generated shim already clears the same class of variables for version inspection. Focused fix: #156.

4. **The dirty-package cache assertions are disabled by the fixture.** [`test/reshim.bats:7-17`](https://github.com/KnickKnackLabs/shiv/blob/dc2f320011fe9b7fa2e0a868ad9c12b7a81d4af7/test/reshim.bats#L7-L17) exports `SHIV_SKIP_CACHE=1`, but [`test/reshim.bats:99-117`](https://github.com/KnickKnackLabs/shiv/blob/dc2f320011fe9b7fa2e0a868ad9c12b7a81d4af7/test/reshim.bats#L99-L117) then treats unchanged cache files as evidence that the dirty guard protected every generated artifact. Those two assertions pass even if cache generation runs before the guard. The shim and alias checks still cover part of the safety claim, but not the cache boundary the test names. Focused fix: #157.

## Adversarial checks

I checked dirty-worktree handling, aggregate exit status, malformed/missing/non-Git registry entries, inherited shell/Git environment, cache-discovery failure, stale-cache removal, exact HEAD/branch preservation, inactive package-store isolation, alias regeneration, and whether the BATS fixtures exercise the production task boundary. I also generated a shim with a pre-feature task map and used it to invoke `reshim` at the exact head; the stale shim correctly fell through to the newly available mise task. The implementation performs no checkout, fetch, reset, or ref write, and the existing ref-preservation assertions passed.

Exact-head `mise run test reshim` passed 5/5 and `codebase lint` passed before the adversarial repros. The PR's hosted Linux/macOS test and completion jobs are green. Each fix-it also passed its focused local suite and hosted Linux/macOS CI; #155 additionally passed the 82 related resolve/completion/update tests locally.

Verification was locally Linux-only. I relied on hosted macOS CI for Bash 3/platform coverage and did not run Windows/PowerShell. Local zsh/fish syntax checks were unavailable, but the exact-head hosted completion jobs are green and the findings do not change completion scripts.

One non-blocking residual characteristic remains: refreshes of a clean package are not transactional across shim, caches, and aliases, so an I/O failure or concurrent package mutation can leave a partial refresh. I did not ask these fix-its to widen the feature into a cross-artifact transaction; the requested changes restore truthful status and the stated dirty-input boundary.

